PR #34, opened 2013-09-08
Add a ReadlineMixin class to emulate some of readline behavior

sitaktif:
Here is a mixin class to add common readline shortcuts to Edit widgets.

It contains tests, doc and an example that I put in the examples/ directory. One might argue that the example is a bit too simple to be there - please give me your feedback.

I think it is compatible with the "highlights" but if I understand correctly, the latter is not working at the moment I could not really test it.

Cheers,
Sitaktif

### DIFF
--- a/docs/reference/widget.rst
+++ b/docs/reference/widget.rst
@@ -200,3 +200,10 @@ Terminal
 
 .. autoclass:: Terminal
 
+Widget Mixin Classes
+----------------------
+
+ReadlineMixin
+~~~~~~~~~~~~~
+
+.. autoclass:: ReadlineMixin
--- a/examples/readline_edit.py
+++ b/examples/readline_edit.py
@@ -1,0 +1,76 @@
+#!/usr/bin/python
+#
+# Urwid readline-emulated edit widget example app
+#    Copyright (C) 2010  aszlig
+#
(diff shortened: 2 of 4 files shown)


rr-:
Is this alive?

rr-:
I made this:

https://pypi.python.org/pypi/urwid-readline/

rr-:
I made this:

https://pypi.python.org/pypi/urwid-readline/

tonycpsu:
urwid-readline seems to fill the underlying need here very well, and seems to be actively maintained.  @rr- , would you have interest in migrating urwid-readline into urwid proper?

rr-:
Nope @tonycpsu  urwid development is too stale for me. 40 commits to master in 2018 alone.

tonycpsu:
@rr- , you're not wrong, but I think you'll find that's about to change.  urwid added two new maintainers yesterday (@ulidtko and myself), and we're starting work on clearing the backlog of PRs, with the hopes of reinvigorating the project.  My thought as I was looking through the backlog was that something like urwid-readline would be a good fit as an urwid-builtin, but I certainly respect your decision.

In any event, urwid-readline is in my view a better path forward than this PR, so I'm going to close this out.  Anyone who wants readline functionality should use urwid-readline.

PR #86, opened 2014-10-20
Increase idle emulation delay

tdryer:
~~Reducing~~ Increasing the idle emulation delay from 1/256 seconds to 1/60 seconds
reduces idle CPU usage without appearing to impact UI responsiveness.

### DIFF
--- a/urwid/main_loop.py
+++ b/urwid/main_loop.py
@@ -1033,7 +1033,7 @@ class TwistedEventLoop(object):
     """
     Event loop based on Twisted_
     """
-    _idle_emulation_delay = 1.0/256 # a short time (in seconds)
+    _idle_emulation_delay = 1.0/60  # a short time (in seconds)
 
     def __init__(self, reactor=None, manage_reactor=True):
         """
@@ -1219,7 +1219,7 @@ class AsyncioEventLoop(object):
     """
     _we_started_event_loop = False
 
-    _idle_emulation_delay = 1.0/256  # a short time (in seconds)
+    _idle_emulation_delay = 1.0/60  # a short time (in seconds)
 
     def __init__(self, **kwargs):
         if 'loop' in kwargs:


tdryer:
I took a stab at making `AsyncioEventLoop` use the same approach as `TwistedEventLoop` for idle callbacks:
https://github.com/wardi/urwid/compare/master...tdryer:asyncio-idle

It works, but there's one added limitation. My application uses a library that adds callbacks to the event loop without using urwid. Since these callbacks don't go though `AsyncioEventLoop.alarm` or `AsyncioEventLoop.watch_file`, they don't cause the idle callbacks to be scheduled. This results in events originating from this library not causing the display to be updated. I think `TwistedEventLoop` has the same limitation.

eevee:
That branch looks okay to me.

wardi:
@eevee please go ahead and merge that and any other fixes you think are ok. If you're up for it I can give you the pypi credentials to put out a release too.

eevee:
Whoops.  The aforementioned limitation applies even to the asyncio socket example — it doesn't need or use either `alarm` or `watch_file`.

It seems like, as far as urwid itself is concerned, `enter_idle` really only cares about running after some input arrives.  `hook_event_loop` takes a callback which is called whenever there's new input, so maybe the screen could be redrawn as part of that process instead?  Or would that break an app like a clock that updates itself even when there's no input?  (I'm not sure how those work at the moment.)

I was curious how much `enter_idle` is used publicly and did a quick [github search](https://github.com/search?l=python&q=enter_idle&type=Code&utf8=%E2%9C%93) — there are only 132 Python results, most of which are from vendored copies of urwid.

wardi:
enter_idle is required for update display to be called regularly, which lots of monitoring programs do.

I don't think this is hard. We can get rid of the hacks in TwistedEventLoop and AsyncEventLoop by removing the enter_idle part of the event loop API and instead have the MainLoop register the same callback for all events from the event_loop (for all FD events and timeouts). That callback will loop on select() and process everything that needs to be done all at once, then do the things it would have done in enter_idle before returning to the event loop.

eevee:
That has the same problem mentioned above: if the event loop doesn't have any timeouts or fd events, the idle callbacks never run.  The asyncio example uses asyncio protocols to read from a socket, so it doesn't listen on an fd at all.

wardi:
sockets have fds. just register them through the MainLoop and everything would work.

eevee:
But then you're not using asyncio primitives, which somewhat defeats the point of having an asyncio adapter.  As written, the example never sees a socket; asyncio automatically makes a new protocol object whenever a new client connects, then occasionally calls a method on that object when new data arrives.

wardi:
If you don't want the update to be "magical" We can require explicit use of the update_display call in cases like that.

We can throttle it internally so that if it's been called too recently set a timer to complete the update instead of doing it immediately. Figuring out what "too recently" is is a little tricky though

wardi:
Another option would be to register a callback to update the display any time the topmost canvas is invalidated. That means we'd need to register which canvases were the topmost on each display with some callbacks or something.

eevee:
I think that sounds more like the Right Thing, yeah.

wardi:
I wish there was a way to set a low-priority 0 second timer. At least on twisted a 0 second timer can fire before pending IO callbacks, which isn't what we want.

How about set it for 0 seconds unless it's been set "too recently" then make it longer.

If only all the event loops came with something like on_enter_idle :-)

rndusr:
This seems to be quite difficult to fix.

Until this is fixed properly, can we at least increase `_idle_emulation_delay`?
I've patched it to `1/25` and reduced the idle CPU load from ~10% to ~1%. Still
not ideal but much better.

tonycpsu:
Issue #90 is open to track a longer-term fix to the underlying issue here.  This workaround makes sense, but 1/60 is still pretty frequent given that a vast majority of apps don't need such a high refresh rate.  I think 1/30 is a good compromise until there's a proper fix, and anyone who wants something more frequent can monkey-patch, as other folks are doing to reduce CPU now.

PR #89, opened 2014-11-06
Remove pollinglistwalker

### LINKED ISSUE #37: Constructing a ListBox with just a list of items uses a PollingListWalker
and `PollingListWalker` is apparently deprecated. If using `ListBox` without an explicit `ListWalker` is supposed to be deprecated, then the error message should say so, instead of only mentioning `PollingListWalker`, which doesn't explicitly occur in user code that does this. If this use is not deprecated, then perhaps this should be switched to another `ListWalker`.

[Relevant code](/wardi/urwid/blob/master/urwid/listbox.py#L284)

hkoof:
Removes the PollingListWalker class and raises a ListBoxError when a ListBox is created without passing an object descending from ListWalker class. Despite its name, PollingListWalker is/was _not_ inheriting from the ListWalker class.

Also changes unit tests and example programs to always use a ListWalker object everywhere a ListBox is created.

Fixes #37

### DIFF
--- a/docs/reference/deprecated.rst
+++ b/docs/reference/deprecated.rst
@@ -10,5 +10,3 @@ Deprecated Classes
 .. autoclass:: FixedWidget
 
 .. autoclass:: AttrWrap
-
-.. autoclass:: PollingListWalker
--- a/examples/dialog.py
+++ b/examples/dialog.py
@@ -111,7 +111,7 @@ class DialogDisplay:
 class InputDialogDisplay(DialogDisplay):
     def __init__(self, text, height, width):
         self.edit = urwid.Edit()
-        body = urwid.ListBox([self.edit])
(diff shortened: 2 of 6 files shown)


tonycpsu:
I'm torn on this PR.  I get the intent, but it's a pretty significant change that will likely break some apps/libraries that are passing in lists.  If the user doesn't have any need to update the list, is there a compelling reason to break compatibility just to have them change `urwid.ListBox(items)` to `urwid.ListBox(urwid.SimpleListWalker(items))`?

wardi:
PollingListWalker prevents urwid from caching the ListBox at all, so it needs to be checked or redrawn on every update. That's really bad for performance, we could switch to wrapping plain lists in SimpleListWalker and issue a warning instead of raising an exception. PollingListWalker has been deprecated for a very very long time

PR #92, opened 2014-11-19
Re-raise exceptions in AsyncioEventLoop properly

rndusr:
Some exceptions (at least UnicodeDecodeError) expect more than 1 argument.
This provides the new exception with all arguments of the original exception.

### DIFF
--- a/urwid/main_loop.py
+++ b/urwid/main_loop.py
@@ -28,6 +28,7 @@ import select
 import os
 from functools import wraps
 from weakref import WeakKeyDictionary
+import sys
 
 try:
     import fcntl
@@ -42,6 +43,33 @@ from urwid import signals
 from urwid.display_common import INPUT_DESCRIPTORS_CHANGED
 
 PIPE_BUFFER_READ_SIZE = 4096 # can expect this much on Linux, so try for that
+
+class ExceptionStorageMixin(object):
+    """Store and raise exceptions in py2 and py3 properly"""
+
+    def _store_exc(self):
+        exc_info = sys.exc_info()
+        if PYTHON3:
+            # Store the exception instance
+            self._exc_info = exc_info[1]
+        else:
+            # Store the exception type, value and traceback
+            self._exc_info = exc_info
+
+    def _maybe_raise_stored_exc(self):
+        try:
+            exc = self._exc_info
+        except AttributeError:
+            pass
+        else:
+            delattr(self, '_exc_info')
+            if PYTHON3:
+                # Simply re-raise previously stored exception
+                raise exc
+            else:
+                # Create new exception from stored type, value and traceback
+                raise exc[0], exc[1], exc[2]
+
 
 class ExitMainLoop(Exception):
     """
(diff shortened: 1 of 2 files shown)


eevee:
Whoops, didn't realize this had fallen through the cracks.  I think this looks reasonable, fwiw.

rndusr:
I've actually used a different solution that I didn't commit because I wanted
to wait a bit longer. And then I forgot to commit it.

Using the PYTHON3 variable works fine for me and is also a lot clearer in my
opinion.

urwid/main_loop.py
```diff
@@ -1324,7 +1324,10 @@ def run(self):
         self._loop.set_exception_handler(self._exception_handler)
         self._loop.run_forever()
         if self._exc_info:
-            raise self._exc_info[0](self._exc_info[1]).with_traceback(self._exc_info[2])
+            if PYTHON3:
+                raise self._exc_info
+            else:
+                raise self._exc_info[0](self._exc_info[1]).with_traceback(self._exc_info[2])
```
eevee: Do you have this `if` backwards?  `with_traceback` only exists in Python 3, but this code will only call it for Python 2.

----

I don't understand what the problem was with the original solution from commit 6c719136:

```
raise self._exc_info[1], None, self._exc_info[2]
```

That's "raise value with traceback" in Python 2, which 2to3 will translate to:

```
raise self._exc_info[1].with_traceback(self._exc_info[2])
```

Which reassigns the exception's traceback (a no-op, since it's the same traceback) and throws it as normal.

The followup commit says "sometimes `sys.exc_info()` returns all Nones", but that sets off all kinds of alarm bells for me and seems like something worth investigating rather than hacking around.

rndusr: Okay, I give up. I can't fix this without proper test cases and all tests are
broken in py3 because of issue #110, which I don't understand (what the hell
is '__super'?), so I'm stuck.

wardi: it's time we remove __super. That was a hack from a really old recipe that lets python2 users get something like the super() in python3.  It's well past its best-before date

eevee: `__super` works fine, but Python 3 in general won't (necessarily) work on a fresh checkout, because urwid still relies on 2to3.  `__super` is set by a metaclass, which is only set by `__metaclass__`, which is just a regular class attribute in Python 3.

Run the tests with `python3 setup.py test`.  That'll do an sdist build (which will run 2to3) and run the tests from within the build.

rndusr: How to remove `__super`? It can't be as simple as replacing every
`self.__super` with `super(<CLASS>, self)`, can it?
If that is really all there is to it, I'm on it.

rndusr:
Ok, I'm hesitantly optimistic that I've got it solved now. At least all tests pass for me.

The culprit was UnicodeDecodeError, which wants exactly 5 arguments, so the `raise...with_traceback()` method failed.

rndusr:
What's the status here?

wardi:
I'll re-run the tests to see if they are passing now

hadware:
I need this, and I have the same bug. Could this be merged in please?

rndusr:
I can no longer reproduce the initial problem. It seems to have been fixed in master.

#163 and #235 might also be fixed.

PR #101, opened 2014-12-21
oddities when rebuilding pngs

bukzor:
This isn't actually a pull request, but the easiest way to show my results.

When rebuilding the doc pngs, I got the below results.
Some came out find, but quite a few are blank, and some are slightly wrong.
Do you have any clues what I'm doing wrong?
Or, is there some regression that happened between the last regeneration and now?
A neat integreation test suite of sorts would be to rebuild the pngs during CI and assert there's no pixel-diffs.


ulidtko:
So, for the most part, this is a rather simple race. See commit messages (coming shortly).

PR #111, opened 2015-01-29
Python 2 and 3 compatibility using six

### LINKED ISSUE #110: attribute '_SelectableIcon__super' missing
On OSX with Python 3.4.2:

$ python3.4 -c "import urwid"

```
  File "<string>", line 1, in <module>
  File "$path/urwid/__init__.py", line 36, in <module>
    from urwid.wimp import (SelectableIcon, CheckBoxError, CheckBox, RadioButton,
  File "$path/urwid/wimp.py", line 99, in <module>
    class CheckBox(WidgetWrap):
  File "$path/urwid/wimp.py", line 104, in CheckBox
    True: SelectableIcon("[X]"),
  File "$path/urwid/wimp.py", line 47, in __init__
    self.__super.__init__(text)
AttributeError: 'SelectableIcon' object has no attribute '_SelectableIcon__super'
```

ttanner:
also fixes #110


nyov:
:+1:
@wardi, can this be merged? Or is the `six` dependency not liked?

wardi:
dropping the "u" from the string literals in the fonts module looks suspect. Doesn't that break the font example in python 2?

Adopting six is fine. We should probably go all the way and replace the existing urwid compat stuff too, but that can come later

ttanner:
can support for python 3.0-3.2 be dropped? Python >=3.3 support u'' again

nyov:
Hm, also using print function without importing it for Python2. Doesn't seem well-tested on second look?
For portable unicode, six has [`six.text_type(str)`](https://pythonhosted.org/six/#six.text_type).

ttanner:
The file urwid/urwid/font.py might be a problem. There is no way to specify unicode raw literals in both Python 2&3, see http://matplotlib.org/devel/portable_code.html
The only solution is escape the \ in the strings.
However, I did not notice any problems with just r"" for the respective literals on both Python 2.7 and 3.4.
"from __future__ import unicode_literals" is supported since Python 2.6 and could be used to stay Python 3.0-3.2 compatible.

@nyov on Python 2 "print(x)" is the same as "print x" if x is a single value, which is the case in all my modifications.

rdsalemi:
I'm running into this bug at this point.  Is there any progress on fixing the issue or a workaround?

rdsalemi:
The anaconda release of urwid seems to fix this problem.

and3rson:
We've finally made the lib 2&3-compatible. This PR seems old enough & replicates some refactoring for Python 3.x support that has recently been merged. I'm closing it.

PR #116, opened 2015-03-19
Change SelectableIcon default cursor_position to 0.

### LINKED ISSUE #115: SelectableIcon should focus first character by default
`SelectableIcon` widget is current having `cursor_position = 1` by default which would result in:

```
S[o]mething
```

with the cursor on `o`.

Commonly the cursor is either at be first character (0) or at the end of the line. I'd suggest to make this default either 0 or -1 and support -1 for last character.

wernight:
Fixes #115. Most users want to have it on the first character, so it should be
the default as well.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,6 +24,8 @@ Urwid 1.3.0
 
  * Documentation fixes (by Ismail, Matthew Mosesohn)
 
+ * SelectableIcon using cursor_position=0 by default instead of 1.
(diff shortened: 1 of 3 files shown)


wardi:
default was set to 1 for CheckBox, RadioButton. I don't see those fixed in this PR

and3rson:
Sorry, totally forgot to validate that. Good catch, I've fixed it.

PR #117, opened 2015-03-20
Fixed capitalisation of "GitHub"

Hoolean:


wardi:
This is a generated file. The source of this document is in https://github.com/wardi/urwid/blob/master/docs/tools/templates/indexcontent.html

PR #120, opened 2015-04-12
Add bracketed paste mode escape sequences

tdryer:
Add [bracketed paste mode](https://cirw.in/blog/bracketed-paste) escape sequences to `escape.py` so urwid can pass these sequences to applications as "begin paste" and "end paste" keys.


wardi:
@tdryer thanks!

PR #121, opened 2015-04-13
remove duplicate Columns.dividechars declaration

uSpike:
Seems to be a typo, dividechars is set twice in `Columns.__init__`


and3rson:
Thanks!

Issue #83, opened 2014-10-12
Italicizing Text


ids1024:
Urwid allows for making text bold or underlined, but as far as I can tell does not allow for italics. Is it possible to make text italicized in urwid and if not, could it be added in a future version?

I apologize if I'm missing something obvious.

wardi:
AFAIK italic text isn't supported in most terminals I'm aware of. closing this issue for now, but please reopen if you know a way to do that.

ids1024:
Quick tests show that italics work in urxvt, st, xterm, and gnome-terminal.  I presume many others support italics too.  It may not be universal, but it is hardly an obscure feature.  I tested with this command:

```
echo "\e[3mHello World\e[23m"
```

wardi:
That's great! we should absolutely add support for italics

lengstrom:
I'm somewhat interested in helping add italics/bold to this project - how do you recommend starting out doing so?

ids1024:
@lengstrom I've already written a pull request (#122) but it has not been merged. It only works in `raw_display`, and the same funtionality should be added to `curses_display` if possible (it may not be), and it needs to be documented.

wardi:
@ids1024 thanks, I've merged your PR

PR #122, opened 2015-04-19
Implement italics support

ids1024:
Italics support, as I requested in issue #83.

Only implemented with raw_display (it might be possible to implement curses_display, but there is not curses.A_ITALICS to match curses.A_BOLD).

It also needs to be added to the documentation, which I have not done.


wardi:
Sorry for letting that sit. If you have a chance please update the docs to mention this new feature.

PR #124, opened 2015-04-27
Python3 fix in graph example

KennethNielsen:
Hallo urwid devs and thanks for a cool library

I encountered a minor python 2 vs 3 incompatibility in the graph.py example. There is a range that is multiplied with 2, but in python 3 range returns a range object which is a iterator or sorts and it doesn't like to get multiplied. Fixed by wrapping list around the range in the one place where it happens.

NOTE: For efficiency, one might considered replacing range with xrange in python2;

``` python
try:
    range = xrange
except NameError:
    pass

```

but with only ever calling range on something like 100, I really don't think it is worth it..

### DIFF
--- a/examples/graph.py
+++ b/examples/graph.py
@@ -48,7 +48,7 @@ class GraphModel:
     data_max_value = 100
 
     def __init__(self):
-        data = [ ('Saw', range(0,100,2)*2),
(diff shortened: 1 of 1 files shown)


and3rson:
Good catch. Thanks!

PR #125, opened 2015-04-27
Browse example, python 3 exception and print fix

KennethNielsen:
A minor fix for python 3 in the browse example.

It just changes an exception handling syntax and uses the print function instead of print statement.

I'm not sure this is all that is required for proper python 3 support, at least, when skimming the code there was a: `somestring is '/'` that cought my eye as potentially problematic, depending on the type of somestring, but I have not investigated this.

### DIFF
--- a/examples/browse.py
+++ b/examples/browse.py
@@ -31,6 +31,8 @@ Features:
 - outputs a quoted list of files and directories "selected" on exit
 """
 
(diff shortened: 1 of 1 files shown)


and3rson:
I'm closing this because the library is now fully ported to Python 3.

Issue #127, opened 2015-06-24
Issues with urwid.ExitMainLoop()


winbornejw:
When exiting the main loop with the latest version available in pip, raising ExitMainLoop doesn't look to be existing cleanly or completely until the entire script is closed. I noticed this because my tool is designed to run bash commands after urwid's work is done. I removed as much as possible to show the issue here. This is urwid 1.3.0

    import urwid
    import os
    
    
    class TicketLoginUI(urwid.WidgetPlaceholder):
        def __init__(self):
            self.LastTicketNumber = None
            self.menu()
    
        def menu(self):
            choices = u'Chapman Cleese Gilliam Idle Jones Palin'.split()
    
            footer = urwid.Columns([
                (9, urwid.Button('Close', on_press=self.ChooseTicket, user_data=None)),
            ])
    
            choicelist = []
            for choice in choices:
                button = urwid.Columns([
                    (24, urwid.LineBox(urwid.Button(choice, self.ChooseTicket, choice))),
                ])
                choicelist.append(urwid.AttrMap(button, None, focus_map='reversed'))
    
            linebox = urwid.LineBox(urwid.ListBox(urwid.SimpleFocusListWalker(choicelist)))
    
            body = [('weight', 1, linebox), ('pack', footer)]
    
            self.original_widget = urwid.Pile(body)
    
        def ChooseTicket(self, button, ticketnumber=None):
            self.LastTicketNumber = ticketnumber
            raise urwid.ExitMainLoop()
    
    
    ticketloginui = TicketLoginUI()
    urwid.MainLoop(ticketloginui, palette=[('reversed', 'standout', '')]).run()
    if ticketloginui.LastTicketNumber:
        command = "echo 'You chose {choice}.'; sleep 10".format(choice=ticketloginui.LastTicketNumber)
        os.system(command)

winbornejw:
I did some more poking and found that this is also an issue in 1.2.2 but not 1.1.3.

winbornejw:
Been installing many versions of urwid and problem seems to have started between 1.2.0 and 1.2.1.
1.2.0 works fine and 1.2.1 has the issue.

wardi:
is it broken on master? Are you using git bisect to find out when it broke?

winbornejw:
It is giving the issue with master as well and I was building specific commits using pip to test them. Can't say I'm familiar with git bisect.

wardi:
@winbornejw look it up, you'll love it for this sort of thing!

[...]

PR #128, opened 2015-06-29
fix for #127 Added a \r for the screen to be properly restored after …

winbornejw:
…screen.stop()


wardi:
Thanks! This CI failure just looks like a new twisted/python 3.2 incompatibility.

PR #137, opened 2015-07-15
minor compat issues

douglas-larocca:
Changed interpreter directives to use `/usr/bin/env python` so running examples will use virtual environment executables.

Minor changes to the subproc.py example for py2/3 cross-compat.


and3rson:
Hashbang update is something really great. I'm going to handle it next week.

Issue #126, opened 2015-04-30
Terminal has echo off when exiting from an exception.


regebro:
The terminal has echo off everytime the program exits because of an exception (Except ExitMailLoop). This means I have to type "reset" all the time, and it's extremely annoying. Workarounds appreciated.

regebro:
I made a workaround like this:

    class MainLoop(urwid.MainLoop):
        def run(self):
            try:
                self._run()
            except urwid.ExitMainLoop:
                pass
            except Exception:
                self.screen.stop()
                raise

I think this probably is how urwid.ExitMainLoop.run() should look. I did see a documentation claiming that this is what it does, but it doesn't.

ngzhian:
@regebro :+1:  for this issue. I have a habit of using `Ctrl+C` to exit my programs, this causes the same issue as well! Do you want to work on a PR for this, I think @wardi may appreciate that. Else I have some time to just spin up a PR based on the code you've already written.

wardi:
reported fixed by #149

PR #149, opened 2015-09-06
Reset screen to original state on any exception.

rianhunter:
In MainLoop._run() we were running self.event_loop.run() in
an exception frame so that we could reset the screen in the
case of an unclean shutdown of the MainLoop. The catch specification
was "except Exception as e" but this doesn't include exceptions that
don't derive from Exception, like KeyboardInterrupt (i.e. Ctrl-C).
This change fixes that.

### DIFF
--- a/urwid/main_loop.py
+++ b/urwid/main_loop.py
@@ -374,9 +374,9 @@ class MainLoop(object):
 
         try:
             self.event_loop.run()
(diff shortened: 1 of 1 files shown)


wardi:
Thanks!

hkoof:
Nice. This fixes #126 too.

PR #151, opened 2015-09-15
Fix default screen size on raw display

inducer:


and3rson:
Makes sense, thanks!

PR #154, opened 2015-09-23
Implement rudimentary support for 'next selectable' for Pile()/Columns()

rianhunter:
Lmk if this is a good general approach!

### DIFF
--- a/urwid/container.py
+++ b/urwid/container.py
@@ -103,6 +103,18 @@ class WidgetContainerMixin(object):
                 return out
             out.append(w)
 
+    def focus_first_selectable(self):
+        """
+        Set the focus of the container to the selectable child widget
+        that comes first in its hierarchy.
+        """
+        for j in range(0, len(self.contents)):
+            if not self.contents[j][0].selectable():
+                continue
+
+            self.focus_position = j
+            return
+
 class WidgetContainerListContentsMixin(object):
     """
     Mixin class for widget containers whose positions are indexes into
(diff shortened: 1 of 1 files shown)


and3rson:
@wardi Thoughts on this? This sounds like a good feature.

urwid/container.py
```diff
@@ -2282,6 +2302,14 @@ def keypress(self, size, key):
 
             self.focus_position = j
             return
+
+        if self._command_map[key] == 'next selectable':
+            # we wanted the next selectable but we were already at the end
+            # set focus back to first selectable
+            # (but still return key to parent in case we should traverse
+            #  to sibling)
+            self.focus_first_selectable()
```
wardi: changing the focus *and* sending the key back to the parent is a behaviour unlike any other widget. I don't think it makes sense to wrap around inside this container. Maybe let an outer container force wrapping if that's desired (more flexible, allows wrapping through multiple child widgets etc.)

urwid/container.py
```diff
@@ -103,6 +103,18 @@ def get_focus_widgets(self):
                 return out
             out.append(w)
 
+    def focus_first_selectable(self):
+        """
+        Set the focus of the container to the selectable child widget
+        that comes first in its hierarchy.
+        """
+        for j in range(0, len(self.contents)):
```
wardi: positions aren't always integers, so it would be better to start from the current focus and get the next child until you find a widget that is selectable. You'll also want to give up after some fixed number because containers may also have an unbounded number of widgets (and might be reading widgets off network or disk when they are iterated over)

geier:
I can re-offer the approach taken in khal: https://github.com/pimutils/khal/commit/bd7c5f928a7670de9afae5657e66c6dc846688ac

PR #160, opened 2015-11-21
python3 compatible function definitions

nyov:


and3rson:
I believe this is old enough & also duplicates some more recent porting of code to Python 3. I'm currently closing this.

PR #161, opened 2015-11-21
python3 compatible exception handling

nyov:

PR #162, opened 2015-11-21
python3 compat

nyov:
#111 cleaned up + #160 + #161 + py3 compatible examples.

(just noticed this also closes #124,  #125)

So I believe this should actually make it py3 compatible now, for the most part


.travis.yml
```diff
@@ -4,6 +4,7 @@ python:
   - "2.7_with_system_site_packages"
   - "3.2_with_system_site_packages"
   - "3.3"
+  - "3.4"
```
ids1024: Why not 3.5 as well?

nyov: Probably it hasn't been tested back then. I just 'merged' the previous efforts.
And I haven't tested with 3.5 myself, so can't rightly add it here.

nyov:
Actually this shouldn't be merged as long as 2to3 conversion happens in setup.
2to3 will mess up things. So more work required in regards to py3 unicode.
While I did some testing there, I didn't quite manage to find a proper solution.

and3rson:
I believe this is old enough & also duplicates some more recent porting of code to Python 3. I'm currently closing this.

PR #167, opened 2015-12-29
indexcontent.html: Fix links after move to urwid/urwid

techtonik:


wardi:
Nice! Thanks, I'll rebuild the docs shortly

PR #175, opened 2016-02-14
Correct error messages in Filler

rr-:
The error message talks about conditions for `fixed bottom` argument to be valid, when the argument is set to `fixed top`. This doesn't seem right.


and3rson:
This makes sense. Thanks!

PR #176, opened 2016-02-14
Correct Edit documentation

rr-:
There were missing doc comments for a few basic properties, + the text about `edit_text` property being a read-only property was a lie.


and3rson:
This makes sense. Thanks!

PR #178, opened 2016-03-09
Add UTF-8 boundary split test

pquentin:
This ensures splitting is done on code points, not bytes.

I was investigating an issue with a package that uses urwid, and this ensures the issue is not in urwid. I thought one more test was never a bad idea.


pquentin:
The test is easy to fix, but would not add anything anyway, as the bug I was chasing was elsewhere.

PR #187, opened 2016-04-11
Switch to relative imports of the urwid modules

andy-z:
For my project I cannot install `urwid` in standard `PYTHONPATH` and I cannot easily change `PYTHONPATH` for other clients that use my project.To handle this I'd like to copy `urwid` inside my own project and import it with `import from` syntax, e.g.:

```
from my_project.extras import urwid
```

Right now `urwid` imports its own modules using absolute names (e.g. `from urwid.container import Frame`) which makes it impossible to relocate easily. 

Proposed patch replaces all absolute imports for `urwid` modules in `urwid` source with relative ones. I have tested it to the best of my ability and also ran it through `pylint` to verify that all imports look OK. I'd appreciate if this could be incorporated into one of the next `urwid` versions.

### DIFF
--- a/urwid/__init__.py
+++ b/urwid/__init__.py
@@ -20,59 +20,59 @@
 #
 # Urwid web site: http://excess.org/urwid/
 
-from urwid.version import VERSION, __version__
-from urwid.widget import (FLOW, BOX, FIXED, LEFT, RIGHT, CENTER, TOP, MIDDLE,
+from .version import VERSION, __version__
+from .widget import (FLOW, BOX, FIXED, LEFT, RIGHT, CENTER, TOP, MIDDLE,
     BOTTOM, SPACE, ANY, CLIP, PACK, GIVEN, RELATIVE, RELATIVE_100, WEIGHT,
     WidgetMeta,
     WidgetError, Widget, FlowWidget, BoxWidget, fixed_size, FixedWidget,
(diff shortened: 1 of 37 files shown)


horazont:
Why can’t you put urwid *next* to your project?

andy-z:
By "next" I guess you mean to be in $PYTHONPATH? The answer is that it's mostly politics - our project is humongous with very many components and we want to avoid possible collisions (now or in the future) with other sub-projects.

horazont:
I really don’t feel this is something which should be solved in urwid.

rndusr:
Is there any reason not to use relative imports?

andy-z:
I do not know other way to solve problems like mine (except forking which I have done already), this is why I suggested this patch. In my view relative imports do not have any drawbacks, only benefits.

and3rson:
Using absolute imports is in general a better idea. I'm not sure if relative imports is something we want.
It's generally recommended either to:

1. Install `urwid` as a package
2. Play around with `sys.path` to point at `urwid` location.

I'm currently closing this.

andy-slac:
Sorry to be obtuse but how absolute imports are better than relative? I believe relative imports were introduced to resolve all kinds of issues with absolute imports and Python conventional wisdom says that it's _always_ better to use relative imports.

wardi:
@and3rson :+1: installing urwid as a package means we can ask users "what version is installed" when they run into an issue. That's very useful!

PR #189, opened 2016-05-06
Fix the asyncio example, and make the raw Screen work without real files

eevee:
I have no idea how this worked in the first place — `Screen` tries to call `.fileno()` on its filehandles all over the place, e.g. to get the screen size.  Now it doesn't, which in particular leaves the example `AsyncScreen` stuck at 80×24, but there's not really any helping that without sending the screen size out-of-band.

### DIFF
--- a/examples/asyncio_socket_server.py
+++ b/examples/asyncio_socket_server.py
@@ -6,14 +6,21 @@ even Python 2 if you install `trollius`!
 """
 from __future__ import print_function
 
-import asyncio
+try:
+    import asyncio
+except ImportError:
+    import trollius as asyncio
+
 from datetime import datetime
 import sys
 import weakref
 
 import urwid
 from urwid.raw_display import Screen
+from urwid.display_common import BaseScreen
 
+import logging
+logging.basicConfig()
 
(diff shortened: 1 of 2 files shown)


and3rson:
@eevee Hi, and thanks for the PR! I know this is ancient, but... would you mind to merge upstream into your branch? I'm going to finally put my hands on this :)

tonycpsu:
This PR will need some freshening up to be usable.  For one thing, `async` is a keyword as of Python >=3.5, so `asyncio.async` yield a syntax error.  Meanwhile, the socat example is showing `TypeError: ('data argument must be byte-ish (%r)', <class 'str'>)` for me when receiving data.

eevee:
Hey, I've dusted this off.  Can't promise I entirely remember what I was doing when I first wrote it, though.  :)

This also fixes local echo of escape sequences, by using socat's `rawer` instead of the now-deprecated `raw`.  The only remaining visual anomaly is that some old contents of the alternate buffer might appear at the bottom of the screen, but this is because the server can't possibly know how big the remote terminal is and defaults to 24 rows.

PR #190, opened 2016-05-23
Convert most examples to python3

belak:
This should make most examples work with both python2 and python3. I didn't bother converting the ssh example, as the SSH library used in the example seems to have gone through some major refactors since this was last touched.

### DIFF
--- a/examples/graph.py
+++ b/examples/graph.py
@@ -48,13 +48,13 @@ class GraphModel:
     data_max_value = 100
 
     def __init__(self):
-        data = [ ('Saw', range(0,100,2)*2),
+        data = [ ('Saw', list(range(0,100,2))*2),
             ('Square', [0]*30 + [100]*30),
-            ('Sine 1', [sin100(x) for x in range(100)] ),
+            ('Sine 1', [sin100(x) for x in list(range(100))] ),
             ('Sine 2', [(sin100(x) + sin100(x*2))/2
-                for x in range(100)] ),
+                for x in list(range(100))] ),
             ('Sine 3', [(sin100(x) + sin100(x*3))/2
-                for x in range(100)] ),
+                for x in list(range(100))] ),
             ]
         self.modes = []
         self.data = {}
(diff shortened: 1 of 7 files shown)


examples/graph.py
```diff
@@ -48,13 +48,13 @@ class GraphModel:
...
+        data = [ ('Saw', list(range(0,100,2))*2),
             ('Square', [0]*30 + [100]*30),
-            ('Sine 1', [sin100(x) for x in range(100)] ),
+            ('Sine 1', [sin100(x) for x in list(range(100))] ),
             ('Sine 2', [(sin100(x) + sin100(x*2))/2
-                for x in range(100)] ),
+                for x in list(range(100))] ),
             ('Sine 3', [(sin100(x) + sin100(x*3))/2
-                for x in range(100)] ),
+                for x in list(range(100))] ),
```
wardi: these three `list()`s aren't necessary because the generators are inside a list comprehension

examples/subproc2.py
```diff
@@ -3,6 +3,6 @@
 import sys
 
 num = int(sys.argv[1])
-for c in xrange(1,10000000):
+for c in range(1,10000000):
```
wardi: really, really don't want this in python2 :-) maybe an explicit check for PYTHON2 is in order

belak:
I'm not very familiar with it, but maybe there's some stuff in six for the stuff you object to.

bwesterb:
`from six import range`

examples/subproc2.py
hobarrera: It's really an example, does this really matter?

and3rson:
@belak You can now use `xrange` from `urwid.compat` for this.
I'll handle this tomorrow.

and3rson:
I just went through and most of this was already done within the scope of `python-dual-support`. Thanks!
Sorry for discarding your work, it's been a long time since the code was maintained.

PR #197, opened 2016-07-13
Add 'linefocus' option to Columns class

rndusr:
This makes it possible to render a whole line as focused with different display attributes for each widget.

### DIFF
--- a/urwid/container.py
+++ b/urwid/container.py
@@ -1699,12 +1699,14 @@ class Columns(Widget, WidgetContainerMixin, WidgetContainerListContentsMixin):
     _sizing = frozenset([FLOW, BOX])
 
     def __init__(self, widget_list, dividechars=0, focus_column=None,
-        min_width=1, box_columns=None):
+        linefocus=False, min_width=1, box_columns=None):
         """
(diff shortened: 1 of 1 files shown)


topisani:
+1. Id be really happy for this one!

wardi:
if you're trying to highlight a whole row just wrap your columns with an AttrMap instead of putting the AttrMap inside each column

rndusr:
But that doesn't work if I want the widgets in the Columns to have different
display attributes.

Take this file list for example:

    File name    File size
    foo.jpg      421kB
    bar.txt        2kB

I want the name to be dark blue on black and the size to be dark green on
black. Focused lines are light blue on dark gray and light green on dark gray.

wardi:
sure it does.

```python
AttrMap(
    Columns([
        AttrMap(Text('foo.jpg'), 'filename'),
        AttrMap(Text('421kB'), 'filesize')]),
    None,
    {'filename': 'filename-focus', 'filesize': 'filesize-focus'})
```

rndusr:
Oh. Okay. I'll try that.
Thanks for your help.

PR #201, opened 2016-08-15
fix iterator in WidgetContainerListContentsMixin

marlox-ouda:
(x)`range` is an iterable
`__iterator__` should return an iterator (`iter(iterable)`)


wardi:
That looks right, these build errors seem to be unrelated. Would be great to start pinning the dependencies so our tests don't go stale. @marlox-ouda do you have time to fix our tests so we can see that this works on all the pythons urwid supports?

marlox-ouda:
Now the tests pass. :)

PR #202, opened 2016-08-16
 Fix travis

marlox-ouda:
Done
* Fix tests for python version 2.6, 2.7, 3.3 and pypy (by removing --use-mirrors option in pip)
* Add successful tests for python version 3.4 and 3.5

[Edit: the following points are no more relevant]
Issue
* Test for python version 3.2 still fails.

I propose to remove the official support for python version 3.2
* tornado stops 3.2 support after tornado version 4.3
* twisted had never officially supported python 3.2.
pip provides the following (recent) versions of twisted : 15.0.0, 15.1.0, 15.2.0, 15.2.1, 15.3.0, 15.4.0, 15.5.0, 16.0.0, 16.1.0, 16.1.1, 16.2.0, 16.3.0, 16.3.1.
python 3 is supported since twisted 15.3.0 and none of the following versions show python3.2 according the provided categories.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,9 @@ before_install:
   - "sudo apt-get update"
   - "sudo apt-get install python-gi python3-gi"
 install:
-  - "pip install setuptools tornado"
+  - "pip install setuptools"
   - if [[ $TRAVIS_PYTHON_VERSION == 2.6 ]]; then pip install "twisted<=15.4"; fi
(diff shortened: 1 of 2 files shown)


marlox-ouda:
Finally, I had done a mistake with the imported version of tornado.

Python v3.2 is now supported again in Travis !

wardi:
@marlox-ouda This is great! Thank you!

PR #207, opened 2016-11-11
Fix missing `modified` signal in SimpleFocusListWalker

zrax:


DamianAngelone:
Will the 'modified' signal for SimpleFocusListWalker be added anytime soon?

wardi:
I wrote this so long ago that I can't remember, but I think the modified signal is for the contents of the list not the focus setting. What does this change fix for you?

zrax:
It's been a while since I submitted this as well...  I believe `SimpleListWalker` and `TreeWalker` both emit 'modified' when the selection changes, so I added it here too.  Either way, we need a signal to indicate when the focus has changed, in case other things need to be updated when the selection changes.

PR #213, opened 2017-01-10
FloatEdit class to edit floats like the IntEdit class to edit integers

hootnot:
class FloatEdit added to be able to edit floating point numbers


tomasfrtala:
Thank you @hootnot it works perfectly! Waiting to see it in upstream.

wardi:
when treating user input as floats developers and users may run into decimal floating point rounding errors, what about using the Decimal class instead to accept exact decimal values without floating point errors?

and3rson:
I agree, would definitely be better to use `Decimal` for better UX here.

urwid/widget.py
```diff
@@ -1733,6 +1733,71 @@ def value(self):
...
+        if not unhandled:
+        # trim leading zeros
+            while self.edit_pos > 0 and self.edit_text.count('.') > 1 and \
+                  self.edit_text[-1] == ".":
+                self.set_edit_pos(self.edit_pos - 1)
+                self.set_edit_text(self.edit_text[:-1])
+
+        return unhandled
+
+    def value(self):
```
and3rson: 1) I'd name it `get_value`
2) Would be cool to add an extra keyword arg `as_decimal=False` or `as_str=False` and return `Decimal` or `str` when this kwarg is `True` instead of casting it to float.

hootnot: Personally I don't like the approach of different methods to get the value.
But it is an option to do it this way, which will give the flexibility of different types and still one single method to get the value (or by property)

```python
from decimal import Decimal

class FloatEdit(object):
    def __init__(self, v, rtype=float):
        self._v = v
        self._rtype = rtype

    @property
    def v(self):
        return self._rtype(self._v)

    def get_value(self):
        return self._rtype(self._v)


fef = FloatEdit(3.14)   # floatedit float
print(fef.get_value())   # or
print(fef.v)

fed = FloatEdit(3.14, Decimal) # floatedit Decimal
print(fed.get_value())
print(fef.v)

fes = FloatEdit(3.14, str)  # floatedit str
print(fes.get_value())
print(fef.v)

```

urwid/widget.py
```diff
@@ -1733,6 +1733,71 @@ def value(self):
...
+        Return the numeric value of self.edit_text.
+        >>> e, size = FloatEdit(), (10,)
+        >>> e.keypress(size, '5')
+        >>> e.keypress(size, '1')
+        >>> e.keypress(size, '.')
+        >>> e.keypress(size, '5')
+        >>> e.keypress(size, '1')
+        >>> e.value() == 51.51
+        True
+        """
```
hootnot: I think returning 0 in case self.edit_text yields False is not OK. A value of 0 is a valid integer (or float), which is definitely something else than an unedited value. This goes for the IntEdit class too. It think it is better to  raise a ValueError, so I would suggest:

```python
if self.edit_text:
    return self._rtype(self.edit_text) 

raise ValueError("no value edited")

File: urwid/widget.py (hunk 1 above)
Comment (wardi) [reply urwid/widget.py]: `rtype=float` looks *very* strange to me. It's simpler, easier to understand, and more explicit (zen of python) for the caller to convert the type, e.g. `float(fed.v)` when they need it.

Internally the only representations that make sense for `self._v` is `str` or `Decimal`. floats shouldn't be allowed to avoid the rounding UX issues. Might even want to raise an error if the user passes a float.

```python
>>> Decimal(0.01)
Decimal('0.01000000000000000020816681711721685132943093776702880859375')
```
would be very confusing to most users

hootnot: my snippet is a bit misinterpreted. Of course is the internal value a string (self.edit_text).

If it is more *Zen*  to convert by the user then I suggest:

  + only *Decimal* is returned, it is up to the user to perform other conversions from the return value if needed
  + input can be *str* or *Decimal*. Checks are performed on the input to check if it is a valid numerical value. If *float()* conversion fails it raises a ValueError
  + the *value()* raises a ValueError in case the default was None and the user did not edit the value. Currently 0 is returned (as IntEdit does) which is definitely not a correct value
  + a parameter *preserveSignificance=True* is added. If the initial value of 1234.0000 is passed and the user edits the 0-'os away to 1234. then the value will still get returned as 1234.0000

if we can agree on this then I will see to make the changes

and3rson: @wardi @hootnot +1 for me. Just keep in mind that decimal mark can be "," or "." in various locales, would be nice to tolerate that as well. Thanks!

and3rson:
Closing in favor of #281.

PR #214, opened 2017-01-21
Emit postchange after text is updated

### LINKED ISSUE #212: Updating frame footer does not work from a `changed` signal handler
This is the simplest program I could find to reproduce the issue.  (Tested on Python 2.7.)

```python
import urwid

def unhandled_input(key):
    if key == 'q':
        raise urwid.ExitMainLoop()
    else:
        if not footer2.keypress((0,), key):
            frame.footer = footer2

def footer2_changed(widget, new_text):
    if not new_text:
        frame.footer = footer1

footer1 = urwid.Text('footer1')
footer2 = urwid.Edit('footer2: ')

urwid.connect_signal(footer2, 'change', footer2_changed)

frame = urwid.Frame(
            urwid.Filler(urwid.Text('body')),
            urwid.Text('header'),
            footer1)

loop = urwid.MainLoop(frame, unhandled_input=unhandled_input)

loop.run()
```
When run, the initial footer is and should be the Text `footer1`.  Then writing some text will switch it out for Edit `footer2: sometext` as intended.  However, deleting the written text (by pressing backspace a few times) should replace `footer2` by `footer1` again, which it doesn't.

Is this indeed a bug or am I abusing the API by changing the layout in a changed handler?

abadger:
Sometimes an Edit's change handler wants to edit the text that has just
been entered.  For instance, if the handler does validation, it may want
to immediately highlight or prevent invalid entries from being shown.
This is not possible if the change signal is emitted before the internal
text data is updated as the update will overwrite anything that the
change handler performed.  Emitting the change event after the update
has been performed allows for changes to be made.

Fixes #212

### DIFF
--- a/urwid/widget.py
+++ b/urwid/widget.py
@@ -1354,10 +1354,10 @@ class Edit(Text):
         """
         text = self._normalize_to_caption(text)
         self.highlight = None
-        self._emit("change", text)
         self._edit_text = text
         if self.edit_pos > len(text):
             self.edit_pos = len(text)
+        self._emit("change", text)
         self._invalidate()
 
     def get_edit_text(self):
--- a/urwid/widget.py
+++ b/urwid/widget.py
@@ -1101,11 +1101,22 @@ class Edit(Text):
     deletion.  A caption may prefix the editing area.  Uses text class
     for text layout.
 
-    Users of this class to listen for ``"change"`` events
-    sent when the value of edit_text changes.  See :func:``connect_signal``.
+    Users of this class may listen for ``"change"`` or ``"postchange"``
+    events.  See :func:``connect_signal``.
+
+    * ``"change"`` is sent just before the value of edit_text changes.
+      It receives the new text as an argument.  Note that ``"change"`` cannot
+      change the text in question as edit_text changes the text afterwards.
+    * ``"postchange"`` is sent after the value of edit_text changes.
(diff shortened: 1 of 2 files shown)


abadger:
Note -- it is possible that some change handlers want to know the pre and post text.  There are several possibilities to support that:

* Introduce a new signal that returns the old text after the change has been committed instead of merely moving the current change signal to after the change has been committed.
* Move the change signal as I have and have the change signal return the old text instead of the new text.
* Move the change signal as I have and havethe change signal return both the old and the new text.

abadger:
After some thought, I've implemented the new signal strategy outlined above.  This way the change signal will remain compatible in case you want to release a 1.3.1.  Code which wants to change the text, needs to use the signal "postchange" instead of the change signal.

abadger:
Found that the checkbox and radiobutton classes' change signal suffers from the same problem.  The latest commit adds a postchange signal to the checkbox class to take care of that.

eonil:
I support this patch.
I don't need the "old value" at all in any case. Especially if you're writing "reactive" UI, such event "before" changing is almost useless, and "after" changing event is far more useful.

Of course, I can get the new text from the handler parameter. But in this case, I need to setup different handler functions for all widgets, and that waste of efforts in "reactive" style UI.

abadger:
Thanks for merging!

PR #216, opened 2017-02-01
support for py3.6

thisch:
I havn't performed any serious testing accept running the unittests locally and the automatic testing by Travis CI.


and3rson:
Thanks! Seems like 3.6 was already added to .travis.yml. I've patched the setup script tags as well.

PR #217, opened 2017-02-19
Support for 24-bit (true color) color attributes.

tonycpsu:
* Interpret colors of the form #rrggbb as 24-bit hex color codes
* Support 24-bit escape sequences
* Update palette_test.py to include demo of 24-bit color support


tonycpsu:
Happy to add docs and such for this if it seems like a good approach to high color support.

tonycpsu:
The test failure there seems unrelated?  Something to do with Twisted's openssl support?

Issue #221, opened 2017-03-16
BUG: raw_display: hook_event_loop does not memoize handles


aleufroy:
I created an application on top of urwid (https://bitbucket.org/aleufroy/lairucrem) which sometime interrupts the main loop to spawn an EDITOR process.

The editor process crashes because `raw_display.Screen.unhook_event_loop` do not "release" the stdout/stdin.

The real issue is introduced by `raw_display.Screen.hook_event_loop` which doesn't save the handles into `self._current_event_loop_handles`.

and3rson:
Waiting on update of #222 for this.

PR #222, opened 2017-03-16
fix: raw display should release file descriptors (issue #221)

aleufroy:
`raw_display.Screen.hook_event_loop` does not fill
`self._current_event_loop_handles` with handled file descriptors.
So, `raw_display.Screen.hook_event_loop` can not release then.


and3rson:
Hi! Thanks for the contribution.

We're now trying to keep urwid alive and are reviewing all open pull requests.

I've recently fixed the build process. In order to properly re-run the build, can you please update your forked repo branch with fresh base upstream to run the build within the latest codebase?

Thanks!

Issue #236, opened 2017-05-30
Strikethrough support


geier:
The [ECMA 48](http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-048.pdf) standard defines an escape sequence for "strikethrough" (search for "crossed-out"). Even though that standard has been released in 1991, for years hardly any terminal supported it (probably because hardly any application made use of it, chicken-egg). But in the last years this has changed and [a lot of free terminal emulators support strikethrough now](https://github.com/pimutils/khal/issues/580#issuecomment-280845290) (also tmux).

I believe urwid should support strikethrough, too.

I have [a branch](https://github.com/geier/urwid/tree/strikethrough) that works for me™, but I'm actually surprised that it does and I don't yet grog what's really going on there, especially with the FG_MASK.

If anybody has time to look at this, I'd be very grateful.

PR #239, opened 2017-06-21
[RFC] New foreground setting "strikethrough"

geier:
After looking through the code again, I'm rather sure that this should be the way to go, see #236.

### DIFF
--- a/docs/manual/displayattributes.rst
+++ b/docs/manual/displayattributes.rst
@@ -169,6 +169,11 @@ Foreground and Background Settings
      - NO
      - NO
      - some support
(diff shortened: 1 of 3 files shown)


and3rson:
Hi! Thanks for the contribution.

We're now trying to keep urwid alive and are reviewing all open pull requests.

I've recently fixed the build process. In order to properly re-run the build, can you please update your forked repo branch with fresh base upstream?

Thanks!

geier:
Done! And thank you for maintaining urwid!

and3rson:
I can confirm that this works. Thanks!

PR #248, opened 2017-08-20
Fix #246, fix #234

### LINKED ISSUE #234: SelectEventLoop bug with alarms?
It seems like `SelectEventLoop` might have a bug in its handling of alarms; in particular, some alarms might get called too late in certain conditions because the heap invariant is violated when an alarm is removed from the heap.

The culprit is in [line 714 of `main_loop.py`](https://github.com/urwid/urwid/blob/master/urwid/main_loop.py#L714) where the alarm is removed from the heap using `self._alarms.pop(0)`. Since `self._alarms` is a heap, simply removing the first element is not guaranteed to maintain the heap invariant; one should use `heapq.heappop()` instead.

The issue becomes apparent only if there are many alarms in the event loop at the same time; violating the heap invariant will mean that some soon-to-expire alarms are "buried" in the middle of the heap instead of floating to the top, so they will not fire until all the other alarms "above" them (with later expiry times) fire as well.

### LINKED ISSUE #246: Unorderable types exception when setting alarm
There's a subtle bug buried SelectEventLoop's `alarm()` method. I've recently encountered this when building an urwid interface for [piwheels](https://github.com/bennuttall/piwheels/blob/master/piwheels/monitor/__init__.py) and while searching for similar instances, I only discovered one (insanum/sncli/issues/35) so it doesn't seem to be commonly encountered.

The issue arises because the `alarm()` method utilizes heapq, but doesn't follow the practice suggested in [the heapq documentation](https://docs.python.org/3/library/heapq.html#priority-queue-implementation-notes) of inserting a tie-break between the time and the callback function. When the alarm entry being inserted has an equal time to something in the current heap, heapq attempts to compare a callback function to another callback function and *if the callback functions are different*, `TypeError` with the message "unorderable types: function() < function()" or "unorderable types: method() < method()" is raised. This is probably why this is a rarely encountered bug: because of the way tuple comparisons work, if the same callback function is always used the issue doesn't occur.

For example:

```python
>>> foo = lambda: 1
>>> bar = lambda: 2
>>> (0.0, foo) < (0.0, foo)
False
>>> (0.0, foo) < (0.0, bar)
  File "<stdin>", line 1, in <module>
TypeError: unorderable types: function() < function()
```

I would suggest following the advice in the heapq documentation (which is probably the simplest fix, although it does use an infinite counter - not necessarily ideal for long-running applications with lots of "fast" alarms). Alternatively, you could store the callbacks in some separate structure which is keyed by a comparable type (like an integer) then pop the callback from that structure when the alarm is retrieved during `_loop` (might be good to address #234 too  given it relates to the same area of the code). I haven't yet checked the other event-loops to see if they're affected but it's probably worth a look. If I get the time this weekend, I'll try and prep a PR for this.

waveform80:
Adjusts the internal representation of an alarm in SelectEventLoop to
include a "tie-break" value prior to the callback (as recommended in the
heapq documentation) to prevent TypeError when comparing two different
callbacks. Also tweaks SelectEventLoop to use heapq.heappop instead of
list.pop for retrieving the next alarm, to maintain the heap structure
of the alarm list.

### DIFF
--- a/urwid/main_loop.py
+++ b/urwid/main_loop.py
@@ -27,6 +27,7 @@ import heapq
 import select
 import os
 from functools import wraps
+from itertools import count
 from weakref import WeakKeyDictionary
 
 try:
@@ -442,7 +443,7 @@ class MainLoop(object):
                 sec = next_alarm[0] - time.time()
                 if sec > 0:
                     break
-                tm, callback = next_alarm
+                tm, tie_break, callback = next_alarm
                 callback()
 
                 if self.event_loop._alarms:
@@ -589,6 +590,7 @@ class SelectEventLoop(object):
         self._watch_files = {}
         self._idle_handle = 0
         self._idle_callbacks = {}
+        self._tie_break = count()
 
     def alarm(self, seconds, callback):
         """
@@ -601,8 +603,9 @@ class SelectEventLoop(object):
         callback -- function to call from event loop
         """
         tm = time.time() + seconds
-        heapq.heappush(self._alarms, (tm, callback))
-        return (tm, callback)
+        handle = (tm, next(self._tie_break), callback)
+        heapq.heappush(self._alarms, handle)
+        return handle
 
     def remove_alarm(self, handle):
         """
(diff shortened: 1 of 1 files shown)


wardi:
@waveform80 thanks!

PR #257, opened 2017-10-20
ListBox: Modify inside docs of `keypress` method.

xndcn:
Since from f5f6711 now `page up` and `page down` will be also handled by widget in focus, so the documents inside the `keypress` method have to be modified to avoid confusing.


and3rson:
Thanks! I've merged this. The builds are now fixed, so I'm ignoring the failed check because this is purely a doc fix.

PR #260, opened 2017-12-24
python 3 compatibility changes

larsks:
this updates the syntax in many of the examples so that they run under
both python 2 and python 3.


wardi:
@larsks thanks!

PR #261, opened 2017-12-28
Fixed issue with Unicode characters in ProgressBar

and3rson:
This commit fixes issues related to Unicode characters in `ProgressBar.get_text` method if overridden.

The problem was that `ProgressBar` tries to insert control characters into bytestring which may lead to the breaking of escape sequences.

For example:
- Char `▶` (`\u25B6`) is encoded into bytestring `'\xe2\x96\xb6'`
- At some point ProgressBar will attempt to split `\xe2` and `\x96` and insert ANSI escape sequence `\x1b[0m` between them to reset the color of the ProgressBar.
- Resulting bytestring will be `\xe2\x1b[0m\x96\xb6` which when decoded will result in `UnicodeDecodeError: [...] invalid continuation byte`

My solution is to slice the original string until the splitting position, encode it into bytes with "ignore" option (to throw away the broken escape sequences) and then decode it back to Unicode string. The resulting string length can be used as a safe position for splitting. This method has a slight overhead, but is very portable and shouldn't provide any significant overhead because ProgressBar content text is usually very short.

Tested in Python 2 and Python 3 with original & extended `ProgressBar.get_text` method.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,11 @@ install:
   - "pip install setuptools"
   - if [[ $TRAVIS_PYTHON_VERSION == 2.6 ]]; then pip install "twisted<=15.4"; fi
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7_with_system_site_packages' ]]; then pip install "twisted==16.6.0"; fi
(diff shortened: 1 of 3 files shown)


and3rson:
Seems like last merged pull requests were ~4 months ago... Is the project still alive?

Issue #267, opened 2018-01-02
Widgets not resizing automatically when using GLibEventLoop


federicotdn:
When using the default main loop, the following example works correctly: the text remains on the center of the screen, even when the terminal is being resized:

```python
import urwid

txt = urwid.Text("Hello World", align='center')
fill = urwid.Filler(txt)
loop = urwid.MainLoop(fill)
loop.run()
```

However, when using `GLibEventLoop` and after resizing the terminal, the text stays in the same spot, until the user presses a key:

```python
import urwid

txt = urwid.Text("Hello World", align='center')
fill = urwid.Filler(txt)
loop = urwid.MainLoop(fill, event_loop=urwid.GLibEventLoop()) # Use GLib event loop
loop.run()
``` 

For context, I'm using:
- Python 3.6.3
- Ubuntu 17.10
- `urwid` version `1.3.1-2ubuntu2`
- `python3-gi` (GObject Introspection) version `3.24.1-2build1`
- Gnome Terminal 3.24.2

federicotdn:
Related: https://github.com/urwid/urwid/issues/25

federicotdn:
It looks like `raw_display.Screen.signal_init()` sets up two signal handlers, one for SIGWINCH and another one for SIGCONT.  The [GNOME documentation](https://developer.gnome.org/glib/stable/glib-UNIX-specific-utilities-and-integration.html#g-unix-signal-source-new) recommends using their own signal handling system instead of the standard UNIX one, when using the `GLib` event loop (the current handlers are not being called when using `GLib`).

I'll try moving the signal handlers setup code to `main_loop.py`, so that each `MainLoop` can setup the signal handlers properly.  The `GLib` signal handlers have different signatures, and must return either `GLib.SOURCE_CONTINUE` or `GLib.SOURCE_REMOVE` to indicate if the signal must be continued to be handled or not, respectively.

federicotdn:
I've been working on top of the `python-dual-support` branch, should I wait until it's merged before I create a PR? @and3rson

and3rson:
@federicotdn I think you can create the PR right away - the diff will automatically list only your new commits once we merge `python-dual-support`.

PR #270, opened 2018-01-05
Allow Screen class to set signal handlers using a function provided by the event loop

federicotdn:
##### Checklist

##### Description:
The Screen class used to set signal handlers using the `signal()` function
from the `signal` standard library module.  This commit changes this
behaviour so that the Screen class now uses a function provided by the active
event loop to set signal event handlers.  This allows us to use GLib
specific signal handling functions when using the GLib event loop.

This PR fixes the second example of issue https://github.com/urwid/urwid/issues/267.

Note: Although the SIGWINCH signal will now by handled correctly, the SIGCONT signal will not be monitored, as the GLib [g_unix_signal_source_new() function](https://developer.gnome.org/glib/stable/glib-UNIX-specific-utilities-and-integration.html#g-unix-signal-source-new) only allows certain signals to be monitored. This is a limitation of GLib.

### DIFF
--- a/urwid/main_loop.py
+++ b/urwid/main_loop.py
@@ -27,6 +27,7 @@ import time
 import heapq
 import select
 import os
+import signal
 from functools import wraps
 from itertools import count
 from weakref import WeakKeyDictionary
@@ -128,6 +129,11 @@ class MainLoop(object):
         if event_loop is None:
             event_loop = SelectEventLoop()
         self.event_loop = event_loop
+
+        if hasattr(self.screen, 'signal_handler_setter'):
+            # Tell the screen what function it must use to set
+            # signal handlers
+            self.screen.signal_handler_setter = self.event_loop.set_signal_handler
 
         self._watch_pipes = {}
 
@@ -645,6 +651,12 @@ class SelectEventLoop(object):
             return True
         return False
 
+    def set_signal_handler(self, signum, handler):
+        """
+        Proxy function to signal.signal()
+        """
+        return signal.signal(signum, handler)
+
     def enter_idle(self, callback):
         """
         Add a callback for entering idle.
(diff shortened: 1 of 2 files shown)


federicotdn:
@and3rson would it be possible to re-run the Travis build? It looks like a test from `TwistedEventLoopTest` failed, and only on Python 3.3, so it doesn't sound like it's related to what I've modified.

and3rson:
This looks interesting. Can you provide a specific example use case for this feature?

and3rson:
Sorry, didn't see the github issue. Makes sense now.

and3rson:
Ideally we should move everything related to signals into main loop classes, but that requires too much refactoring, so this PR is all we need now.
Thanks!

federicotdn:
@and3rson what we have at the moment is a `MainLoop` class, which has two properties: `event_loop`, which can be an instance of `SelectEventLoop`, `GLibEventLoop`, etc. and `screen`, which can be an instance of `raw_display.Screen`, `curses_display.Screen`, `web_display.Screen`, etc.  The `event_loop` and `screen` properties do not have references to each other, and neither do they have references to the `MainLoop` who instanced them.

The problem with this was that a specific `Screen` class (`raw_display`) needed special signal handling when running with a specific event loop (`GLibEventLoop`), so at the end, this `Screen` class needed to be aware of how this event loop wanted to handle UNIX signals.  So my solution was to specify the desired signal handler setter function (such as `signal.signal()`) in each `EventLoop` class, and then pass it to the `screen` property, but only if `screen` is an instance of `raw_display.Screen`, since the other `Screen` classes to not use signals (this was done using `hasattr()`, which is ugly).

and3rson:
@federicotdn thanks for the clarification. Makes perfect sense.

and3rson:
Do you mind adding some docs for this as well to make devs aware of using this feature?

federicotdn:
@and3rson sure, no problem.

PR #271, opened 2018-01-08
Create the EventLoop abstract class, add documentation for signal handling

federicotdn:
##### Checklist

##### Description:
I've created a new abstract class, `EventLoop`, from which all event loop classes now extend.  If any new method or property common to all event loops is needed, it can be added to this class. `EventLoop` also contains the default implementation of `set_signal_handler`, which is only overridden by `GLibEventLoop`.  I've also added some documentation explaining how the signal handling works.

I replaced the old `.gitignore` file by the one provided by GitHub, which I find to be more complete (for example, files created by tox were not being ignored).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,104 @@
-.DS_Store
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
 
-*.pyc
-*.pyo
+# C extensions
+*.so
 
-*.egg-info
-_build
-build
-dist
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
 MANIFEST
 
-.coverage
-coverage
-htmlcov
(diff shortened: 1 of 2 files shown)


and3rson:
Great job! Such refactoring & explicitness is exactly what some components of urwid are missing.
Thank you for this contribution, I'm going to review it asap.

and3rson:
LGTM! @wardi any objections?
P. S. The build fails due to Twisted; seems unrelated to this PR.

wardi:
no objection to the primary change, but that .gitignore seems like overkill. Django configs in urwid? Do people really put virtualenvs inside their cloned git repos? I'd rather I didn't know about that..

and3rson:
I think this gitconfig makes sense, you never know what people put into git repo root :)
Also, yes, I usually put .env inside repo root. Pretty handy.

federicotdn:
@and3rson Ok, so I've installed Python 3.3 using `pyenv`, and I ran the tests again using `tox -e py33`, and all of them passed (I ran them many times).  What's interesting is that, in my previous PR (https://github.com/urwid/urwid/pull/270) the Travis build failed on the same test and Python version.

and3rson:
@federicotdn thanks! I'm going to merge this and later we'll figure out tests issue.

PR #272, opened 2018-01-16
Fix underlining padded text

rndusr:
##### Checklist

##### Description:
When underlining text that ends with spaces, it is only underlined to where the spaces start. This fixes that so the spaces are also underlined.

This demonstrates the issue:

```python
palette = [('deco', 'underline,black', 'yellow')]
txt = urwid.Text('Hello', align='center')
fill = urwid.Filler(urwid.AttrMap(txt, 'deco'))
loop = urwid.MainLoop(fill, palette)
loop.run()
```

Note how the whole line is yellow, but only underlined up to the first space after "Hello".

### DIFF
--- a/urwid/raw_display.py
+++ b/urwid/raw_display.py
@@ -787,7 +787,7 @@ class Screen(BaseScreen, RealTerminal):
                 if (run[-1:] == B(' ') and self.back_color_erase
                         and not using_standout(a)):
                     whitespace_at_end = True
-                    row = row[:-1] + [(a, cs, run.rstrip(B(' ')))]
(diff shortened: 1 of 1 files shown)


and3rson:
Looks reasonable! @wardi thoughts why `rstrip` was there?

wardi:
@and3rson without it urwid sends a huge amount of bytes to the display on every update instead of relying on the clear-to-eol escape sequence. If clear-to-eol doesn't work with underlines we could add a special case for them here instead.

wardi:
@rndusr looks better, thanks

PR #281, opened 2018-02-04
extended numerical editing: integers and floats

hootnot:
##### Checklist

##### Description:
See: #213 , changes implemented as proposed

  + only *Decimal* is returned, it is up to the user to perform other conversions from the return value if needed
  + input can be *str* or *Decimal*. Checks are performed on the input to check if it is a valid numerical value. If *float()* conversion fails it raises a ValueError
  + the *value()* raises a ValueError in case the default was None and the user did not edit the value. Currently 0 is returned (as IntEdit does) which is definitely not a correct value
  + a parameter *preserveSignificance=True* is added. If the initial value of 1234.0000 is passed and the user edits the 0-'os away to 1234. then the value will still get returned as 1234.0000

In a  new module *urwid.numedit*  you find the classes *IntegerEdit* and *FloatEdit*, *PEP8* compliant.
A few changes on the list above:
  +  regarding:  input can be *str* or *Decimal*:  integer values are allowed too
  +  the *value()* method raises a ValueError in case the default was None and the user did not edit the value: the *value()* method returns None in that case, so no exception is raised

The existing *IntEdit* class is untouched. As discussed earlier, it's return value of 0 in case it is un-edited is not OK. I think it should be deprecated in favor of the new *IntegerEdit*.

Apply the patch below to examples/calc.py to use the *FloatEdit* class in the calculator example.

```
diff --git a/examples/calc.py b/examples/calc.py
index 39c37f7..f584cf1 100755
--- a/examples/calc.py
+++ b/examples/calc.py
@@ -36,6 +36,7 @@ from __future__ import print_function
 import urwid
 import urwid.raw_display
 import urwid.web_display
+import urwid.numedit as numedit
 
 # use appropriate Screen class
 if urwid.web_display.is_web_request():
@@ -135,7 +136,8 @@ class Cell:
     def setup_edit(self):
         """Create the standard edit widget for this cell."""
 
-        self.edit = urwid.IntEdit()
+        #self.edit = urwid.FloatEdit(decimalSeparator=",")
+        self.edit = numedit.FloatEdit()
         if not self.is_top:
             self.edit.set_caption( self.op + " " )
         self.edit.set_layout( None, None, CALC_LAYOUT )
@@ -146,7 +148,7 @@ class Cell:
         if self.child is not None:
             return self.child.get_result()
         else:
-            return int("0"+self.edit.edit_text)
+            return float("0"+self.edit.edit_text)
 
     def get_result(self):
         """Return the numeric result of this cell's operation."""
@@ -155,7 +157,7 @@ class Cell:
             return self.get_value()
         if self.result.text == "":
             return None
-        return int(self.result.text)
+        return float(self.result.text)
 
     def set_result(self, result):
         """Set the numeric result for this cell."""
@@ -163,7 +165,7 @@ class Cell:
         if result == None:
             self.result.set_text("")
         else:
-            self.result.set_text( "%d" %result )
+            self.result.set_text( "%.5f" %result )
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,3 +77,4 @@ Contributors
 - `grugq <//github.com/grugq>`_
 - `inducer <//github.com/inducer>`_
 - `winbornejw <//github.com/winbornejw>`_
+- `hootnot <//github.com/hootnot>`_
--- a/docs/reference/widget.rst
+++ b/docs/reference/widget.rst
@@ -68,6 +68,25 @@ TreeWidget
 
 .. autoclass:: TreeWidget
 
+
+.. currentmodule:: urwid.numedit
+
+Extended Numerical Editing Classes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+IntegerEdit
+~~~~~~~~~~~
+
+.. autoclass:: IntegerEdit
+
+FloatEdit
+~~~~~~~~~~~
+
+.. autoclass:: FloatEdit
+
+
+.. currentmodule:: urwid
+
 SelectableIcon
 ~~~~~~~~~~~~~~
 
--- a/urwid/numedit.py
+++ b/urwid/numedit.py
@@ -1,0 +1,230 @@
+# -*- coding: utf-8 -*-
+#
+# Urwid basic widget classes
+#    Copyright (C) 2004-2012  Ian Ward
+#
+#    This library is free software; you can redistribute it and/or
+#    modify it under the terms of the GNU Lesser General Public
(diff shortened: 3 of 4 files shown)


hootnot:
Added in important change to handle leading zero's

It keeps leading 0's in case of non base 10: so editing an 10FF and delete the '1' it becomes 0FF, while on a regular base-10 int, editing 1002 and delete the '1' , it becomes 2.

Guess this is about it

Feite

and3rson:
This looks great, thanks!

PR #298, opened 2018-05-20
[WIP] Trio event loop

abadger:
##### Checklist

##### Description:
Trio is an alternative async framework based around the async and await keywords that were added to recent python versions.  This PR adds an EventLoop class which uses trio as its backend.

I've marked this as a work in progress as async and await are only available in python-3.6 or greater.  I'll have to reorganize the new code so that it is only parsed if we're on python-3.6 or higher.  I also want to check whether the eventloop is being tested in tox and try to get a trio developer to look over my work and make sure I'm not making any big mistakes.

### DIFF
--- a/bin/deps.py
+++ b/bin/deps.py
@@ -22,4 +22,10 @@ try:
 except ImportError:
     pass
 
+try:
+    import trio
+    deps.append("trio")
+except ImportError:
+    pass
+
 print(" ".join(deps))
--- a/tox.ini
+++ b/tox.ini
@@ -12,6 +12,19 @@ usedevelop = true
 deps =
     setuptools
     tornado
+    coverage
+    py27: twisted==16.6.0
+    py33: twisted
+    py34: twisted
+    py35: twisted
+    py36: twisted
+    pypy: twisted
+
+[testenv:py35]
+usedevelop = true
+deps =
+    setuptools
+    tornado
     trio
     coverage
     py27: twisted==16.6.0
@@ -20,6 +33,21 @@ deps =
     py35: twisted
     py36: twisted
     pypy: twisted
+
+[testenv:py36]
+usedevelop = true
+deps =
+    setuptools
+    tornado
+    trio
+    coverage
+    py27: twisted==16.6.0
+    py33: twisted
+    py34: twisted
+    py35: twisted
+    py36: twisted
+    pypy: twisted
+
 commands =
     coverage run ./setup.py test
 
--- a/tox.ini
+++ b/tox.ini
@@ -12,6 +12,7 @@ usedevelop = true
 deps =
     setuptools
     tornado
+    trio
     coverage
     py27: twisted==16.6.0
     py33: twisted
--- a/urwid/__init__.py
+++ b/urwid/__init__.py
@@ -55,7 +55,7 @@ from urwid.command_map import (CommandMap, command_map,
     CURSOR_PAGE_UP, CURSOR_PAGE_DOWN, CURSOR_MAX_LEFT, CURSOR_MAX_RIGHT,
     ACTIVATE)
 from urwid.main_loop import (ExitMainLoop, MainLoop, SelectEventLoop,
-    GLibEventLoop, TornadoEventLoop, AsyncioEventLoop)
+    GLibEventLoop, TornadoEventLoop, AsyncioEventLoop, TrioEventLoop)
 try:
     from urwid.main_loop import TwistedEventLoop
 except ImportError:
--- a/urwid/_async_kw_main_loop.py
+++ b/urwid/_async_kw_main_loop.py
@@ -1,0 +1,172 @@
+#!/usr/bin/python
+#
+# Urwid main loop code using Python-3.5 features (Trio, Curio, etc)
+#    Copyright (C) 2018 Toshio Kuratomi
+#
+#    This library is free software; you can redistribute it and/or
+#    modify it under the terms of the GNU Lesser General Public
+#    License as published by the Free Software Foundation; either
+#    version 2.1 of the License, or (at your option) any later version.
+#
+#    This library is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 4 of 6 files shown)


njsmith:
There is a downside to using `Event` objects to signal cancellation: if you start waiting for something in the background, and then later cancel it, the background task actually keeps running until whatever its waiting for happens. This might be particularly problematic when watching for fds, because if there's a trio task sitting in `wait_readable` and the fd it's waiting for gets closed, you can get weird effects. Like the task might keep sleeping forever, or worse, if another fd is opened and gets assigned the same number as the one that was closed, trio could get confused between them. (And unfortunately, there isn't really anything trio can do about this – it's a limitation of the underlying OS APIs.)

The other option would be to signal cancellation using cancel scopes. 

```python
    async def _trio_watch_file(self, fd, callback):
        with trio.open_cancel_scope() as cancel_scope:
            while True:
                await trio.hazmat.wait_readable(fd)
                callback()
```

Now this task can be cancelled by calling `cancel_scope.cancel()`, and that will immediately clean it up. But there is an awkward problem: how do we get at the `cancel_scope` object from outside `_trio_watch_file`?

Though, really, there is a similar problem with the current code as well: it turns out that currently calling `trio.Event` from outside `trio.run` works, but this isn't guaranteed...

I think maybe something like this should resolve these issues?

```python
##### Helper class

class CancelHandle:
    def __init__(self):
        self._cancelled = False
        self._cancel_scope = None

##### inside TrioEventLoop

def __init__(self):
    # List of (async_fn, args, handle) tuples
    self._queued = []

def _queue_cancellable_operation(self, async_fn, *args):
    handle = CancelHandle()
    self._queued.append((async_fn, args, handle))
    self._maybe_process_queue()
    return handle

def _maybe_process_queue(self):
    if self._nursery is None:
        return
    for (async_fn, args, handle) in self._queued:
        self._nursery.start_soon(self._run_cancellable_operation, async_fn, args, handle)
    self._queued.clear()

async def _run_cancellable_operation(self, async_fn, args, handle):
    # In case we were cancelled before we even started
    if handle._cancelled:
        return
    with trio.open_cancel_scope() as cancel_scope:
        # In case we're cancelled after we started
        handle._cancel_scope = cancel_scope
        try:
            await async_fn(*args)
        finally:
            # to make sure future calls to _cancel_operation return False
            # and to make sure we don't try doing handle.cancel_scope.cancel() after leaving trio
            handle._cancelled = True

def _cancel_operation(self, handle):
    if handle._cancelled:
        return False
    else:
        handle._cancelled = True
        if handle._cancel_scope is not None:
            handle._cancel_scope.cancel()
        return True
```

Then `alarm`, `watch_file`, `enter_idle` become simple wrappers around `_queue_cancellable_operation`, and `remove_*` are all `return self._cancel_operation(handle)`. What do you think?

BTW, I think the current implementation of `enter_idle` is way more complicated than it needs to be :-). Instead of having a task that spawns a task which spawns a task, etc., like you have to in Twisted, you can just write:

```python
async def _idle_loop(self, callback):
    while True:
        await trio.sleep(self._idle_emulation_delay)
        callback()

def enter_idle(self, callback):
    return self._queue_cancellable_operation(self._idle_loop, callback)
```

abadger:
Cool!  How to make use of a cancel_scope outside of its context is exactly
the question I was struggling to find a way to solve.  I'll give that and
the idle changes a try one my plane lands.

njsmith:
Some more thoughts:

- I'm not sure how urwid expects its event loop to respond when a callback raises an unexpected exception, but you should probably check :-)

- currently this doesn't provide any way to run regular trio code alongside urwid code :-). My suggestion: keep the `run` method so you can run urwid's test suite, but don't actually recommend people use it. Instead, provide some methods like:

  ```python
  async def __aenter__(self):
      assert self._nursery is None
      self._nursery = await trio.open_nursery().__aenter__()
      return self

  async def __aexit__(self, *args):
      try:
          await self._nursery.__aexit__(*args)
      finally:
          self._nursery = None
  ```

  So then people could use it like:

  ```python
  async def main():
      async with urwid.TrioMainLoop(...) as urwid_main_loop:
          # arbitrary trio code here

  trio.run(main)
  ```

  I guess you might also want a way for trio code to interact with urwid's main loop exiting logic? Not sure.

- I'm not quite sure what urwid actually expects from an "idle" callback – it's not a very well-defined concept. But trio does actually have a way to wait for everything to be idle: `trio.testing.wait_all_tasks_blocked`. I really can't recommend using this for anything except testing, because it creates such weird global coupling; I think probably sticking with the current "idle" semantics of "we call this callback occasionally" is probably better. But figured I should mention it so this could be an intentional decision.

njsmith:
FYI, there's now an issue on the trio tracker (https://github.com/python-trio/trio/issues/607) for making cancel scopes natively support being created at a different place than they're used.

tonycpsu:
@abadger : is this still something you're working on?

abadger:
Sorry, I've had a lack of time this year, i still have my branch so if
you'd like to close this until i can get more round tuits, that's fine with
me.. (or if you're asking because you'd like to take it over, that's fine
for me too)

tonycpsu:
No worries, I was just going through the backlog and this seemed like something that (a) folks might like and (b) seemed reasonably far along.  We'll just leave it open for now.

ntamas:
I wish I had found this PR earlier because I have just spent a few hours on implementing a Trio event loop for `urwid` on my own. My implementation is to be found in [this gist](https://gist.github.com/ntamas/2e1063a0a7609cf1d34d42c700a02d87). It is reasonably similar to this PR but uses detached cancel scopes as suggested by @njsmith in [this comment](https://github.com/urwid/urwid/pull/298#issuecomment-414528841).

@abadger, do you plan to revive this PR in the near future?

abadger:
@ntamas if you'd like to take over and get your changes merged instead, that suits me just fine :-). This has been a tumultuous couple years for me so i haven't had as much time as if like for working on anything in my free time :-)

ntamas:
Okay, I have sent my version of the Trio-based event loop as PR #298.

PR #330, opened 2018-12-17
urwid.util: Fix bug in rle_append_beginning_modify

BkPHcgQL3V:
rle_append_beginning_modify was not correctly prepending rle fragments with mismatching attributes.

### DIFF
--- a/urwid/util.py
+++ b/urwid/util.py
@@ -301,7 +301,7 @@ def rle_append_beginning_modify(rle, a_r):
         if a == al:
             rle[0] = (a,run+r)
         else:
-            rle[0:0] = [(al, r)]
+            rle[0:0] = [(a, r)]
 
 
 def rle_append_modify(rle, a_r):


tonycpsu:
@BkPHcgQL3V : do you know what the implications of this bug might be?  I see that this function is used by `trim_text_attr_cs`, which in turn is used by `TextCanvas`.  I'm just curious as to how you found this bug in the first place, and whether it might resolve any others.

ulidtko:
This fix is obviously correct, and I've written a straightforward and clear test for it. Same as the OP -- can't predict the consequences exactly, but this code has never worked correctly in the first place. Yes, this might fix a few other bugs (perhaps even the Padding pack behavior #346).

I'll merge this together with the new test (fails on master now, passes with the PR) and a minor local rename `rle_append_beginning_modify` → `rle_prepend_modify`.

mdtrooper:
> This fix is obviously correct, and I've written a straightforward and clear test for it. Same as the OP -- can't predict the consequences exactly, but this code has never worked correctly in the first place. Yes, this might fix a few other bugs (perhaps even the Padding pack behavior #346).
> 
> I'll merge this together with the new test (fails on master now, passes with the PR) and a minor local rename `rle_append_beginning_modify` → `rle_prepend_modify`.

No sorry, it still fails but thanks.

PR #332, opened 2018-12-27
Use asyncio's TimerHandle.cancelled() if available

seleem1337:
##### Checklist

##### Description:
Python 3.7 introduces public `TimerHandle.cancelled()` as seen here: https://bugs.python.org/issue31943

Supporting this public method has a nice side-effect of also supporting `uvloop`, since `uvloop` doesn't expose the private `_cancelled` property.


seleem1337:
It seems the image used in Travis doesn't support Python 3.7, but I'm not familiar enough with Travis to know which image to use.

PR #343, opened 2019-04-15
fix EOF detection for the Terminal widget on Python 3

ntamas:
##### Checklist

##### Description:

This PR fixes the detection of the termination of the spawned process of the Terminal widget on Python 3.x. The problem was that process termination resulted in the `os.read()` call to return `b''` on Python 3.x (instead of `''`), which then failed the comparison with `''`. The PR ensures that we use the correct EOF marker (`b''` on Python 3.x, `''` on Python 2.x).

### DIFF
--- a/urwid/vterm.py
+++ b/urwid/vterm.py
@@ -47,6 +47,7 @@ from urwid.widget import Widget, BOX
 from urwid.display_common import AttrSpec, RealTerminal, _BASIC_COLORS
 from urwid.compat import ord2, chr2, B, bytes, PYTHON3, xrange
 
+EOF = B('')
 ESC = chr(27)
 
 KEY_TRANSLATIONS = {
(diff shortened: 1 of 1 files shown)


ulidtko:
LGTM, even though I cannot reproduce the original problem.

ntamas:
Thanks for the quick response!

For sake of posterity, the original issue can be reproduced like this:

1. Create a Terminal widget and spawn a shell in it.
2. Exit the shell. In Python 3.x, the widget won't notice that the process has exited because `os.read()` returns `b''` and not `''`, so no `closed` signal is emitted.
3. Try typing something into the widget now. urwid will then crash with an `IOError` (at least on macOS where I tested it) because we are trying to write to a file descriptor (the pty) that is no longer valid.

PR #347, opened 2019-05-14
Unbreak python ./examples/treesample HOME END keys.

xnox:
TreeWalker does not implement `.positions` and thus default handler in `__super.keypress` fails and tracebacks trying to process HOME and END keybindings.

However, these keys are implemented in the treelistbox (possibly prior to the generic key handler) and imho it should be used.

With this patch treesample works.


tonycpsu:
Looks good to me.  Thanks!

Issue #317, opened 2018-11-04
Smoothing progress bars doesn't work


rr-:
##### Description:
Progress bars seem to be unable to use the smoothing feature.
Python 3.7.

##### Affected versions (if applicable)
- [ ] `master` branch (specify commit)
- [x] Latest stable version from `pypi`
- [ ] Other (specify source)

##### Steps to reproduce (if applicable)
Create a progress bar, set its progress to miniscule value

##### Expected/actual outcome
It should use the `eighths` variable (`▏▎▍▌▋▊▉`)
It uses only full blocks instead

The reason seem to be in this check:
https://github.com/urwid/urwid/blob/master/urwid/graphics.py#L945

Here, the value of `c._text[0][ccol]` on my machine is `32` (an integer) rather than `" "` (a string).

ntamas:
I can confirm that simply replacing `c._text[0][ccol] == " "` with `ord2(c._text[0][ccol]) == 32` solves the issue and does not break compatibility with Python 2.x. (`ord2` has to be imported from `urwid.compat`).

Note that this is not purely a cosmetic issue; on Python 3.x, if the percentage of the progress bar is so small that the first character would be a block character from `eighths` (a percentage of 1% with a sufficiently narrow progress bar would do), this bug essentially collapses the progress bar to zero width when it is used in a `Columns` widget as parent as we end up in the wrong branch in the rendering method and we render nothing.

ulidtko:
Fix merged, will appear in the next release.

PR #351, opened 2019-08-08
fix ProgressBar smoothing on Python 3.x

ntamas:
##### Checklist

##### Description:

This PR fixes an issue with the "smoothing" feature of the ProgressBar widget on Python 3.x. Due to a bytes-vs-str problem, smoothing essentially doesn't work on Python 3.x as it is because we are indexing into a bytes object (which returns an integer on Python 3.x) and then compare it with a string (`" "`). Using the `ord2` compatibility function from `urwid.compat` and replacing `" "` with its ASCII code 32 fixes the problem.

The PR would fix issue #317.

### DIFF
--- a/urwid/graphics.py
+++ b/urwid/graphics.py
@@ -22,7 +22,7 @@
 
 from __future__ import division, print_function
 
-from urwid.compat import with_metaclass
+from urwid.compat import ord2, with_metaclass
 from urwid.util import decompose_tagmarkup, get_encoding_mode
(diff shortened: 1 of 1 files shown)


ntamas:
Travis CI failure seems to be unrelated; it is a versioning problem with Twisted, which now requires Python 3.5 (and the CI test runs on Python 3.4).

ulidtko:
Tested with `examples/graph.py`, does indeed fix #317. Merging.

PR #354, opened 2019-08-15
listbox kwarg to wrap moving around

kkrolczyk:
##### Checklist

##### Description:
This allows to traverse list, and optionally wrap around, to get back to beginning (or end)

### DIFF
--- a/urwid/listbox.py
+++ b/urwid/listbox.py
@@ -81,11 +81,12 @@ class ListWalker(with_metaclass(signals.MetaSignals, object)):
         except (IndexError, KeyError):
             return None, None
 
-
 class SimpleListWalker(MonitoredList, ListWalker):
-    def __init__(self, contents):
+    def __init__(self, contents, wrap_around=False):
         """
         contents -- list to copy into this object
+
+        wrap_around -- if true, jumps to beginning/end of list on move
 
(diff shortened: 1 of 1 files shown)


urwid/listbox.py
```diff
@@ -82,8 +82,28 @@ def get_prev(self, position):
...
+    def prev_position(self, position):
+        """
+        Return position before start_from.
+        """
+        if position <= 0:
+            if self.wrap_around:
+                return len(self) - 1
+            raise IndexError
+        return position - 1
+
```
ulidtko: Uhhh no, I don't think this belongs here. Even despite that moving these two methods into `ListWalker` deduplicates code from the two descendant classes, `SimpleListWalker` & `SimpleFocusListWalker`.

Two issues:
 1. `ListWalker` is abstract, sort-of like an interface: we can't assume that `position` is of numeric type, and that we can do arithmetic on it. An imaginary `MoonWalker` descendant might be able to use filenames as `position`-s, for example.
 2. `self.wrap_around` is unknown in the scope of `ListWalker`. This will certainly aggravate static analysis tools (pylint, mypy, etc). This will also break other people's custom ListWalker implementations if we merge it.

I'm okay with the `wrap_around` feature — but this little refactoring needs to be undone.

PR #355, opened 2019-08-15
allow setting align for button

kkrolczyk:
##### Checklist

##### Description:
Patch
- separates constants from widget.py to a file and
- lets to set horizontal alignment for button:
'<   button   >'
'<    ✔   >']

### DIFF
--- a/urwid/constants.py
+++ b/urwid/constants.py
@@ -1,0 +1,30 @@
+# define some names for these constants to avoid misspellings in the source
+# and to document the constant strings we are using
+
+# Widget sizing methods
+FLOW = 'flow'
+BOX = 'box'
+FIXED = 'fixed'
+
+# Text alignment modes
+LEFT = 'left'
+RIGHT = 'right'
+CENTER = 'center'
+
+# Filler alignment
+TOP = 'top'
+MIDDLE = 'middle'
+BOTTOM = 'bottom'
+
+# Text wrapping modes
+SPACE = 'space'
+ANY = 'any'
+CLIP = 'clip'
+ELLIPSIS = 'ellipsis'
+
+# Width and Height settings
+PACK = 'pack'
+GIVEN = 'given'
+RELATIVE = 'relative'
+RELATIVE_100 = (RELATIVE, 100)
+WEIGHT = 'weight'
--- a/urwid/constants.py
+++ b/urwid/constants.py
@@ -1,30 +1,44 @@
-# define some names for these constants to avoid misspellings in the source
-# and to document the constant strings we are using
+from enum import Enum, auto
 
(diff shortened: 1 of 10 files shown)


urwid/wimp.py
```diff
@@ -32,21 +32,24 @@
...
     _selectable = True
-    def __init__(self, text, cursor_position=0):
+    def __init__(self, text, cursor_position=0, align=LEFT):
         """
         :param text: markup for this widget; see :class:`Text` for
                      description of text markup
         :param cursor_position: position the cursor will appear in the
                                 text when this widget is in focus
+        :param align: markup for this widget; see :class:`Text` for
+                      description of align
```
ulidtko: Copy-paste whoopsie... align is not markup!

Instead of referring to `:class:Text` — just enumerate the possible values: `'left' (default), 'center', 'right'`, they're pretty self-explanatory

urwid/constants.py
```diff
@@ -0,0 +1,30 @@
...
+ANY = 'any'
+CLIP = 'clip'
+ELLIPSIS = 'ellipsis'
+
+# Width and Height settings
+PACK = 'pack'
+GIVEN = 'given'
+RELATIVE = 'relative'
+RELATIVE_100 = (RELATIVE, 100)
+WEIGHT = 'weight'
```
ulidtko: I think these all need to be made into [`Enum`-s](https://docs.python.org/3.4/library/enum.html).

That'd also allow to actually document the constants — via proper docstrings appearing in Sphinx docs, rather than in-source comments.

kkrolczyk:
Hi @ulidtko i had a while to sit and check it, but TBH it does not look to trivial, actually quite the opposite due to how heavily those constants are embedded in current code. Current solution (from this patch) fails with

```
urwid.container.ColumnsError: initial widget list item invalid: ('fixed', 1, <AttrMap selectable frozenset({<WidgetSize.FLOW: 2>}) widget <SelectableIcon selectable frozenset({<WidgetSize.FLOW: 2>}) widget '-'> attr_map={None: 'dirmark'}>)
```

for examples/browse.py (and probably for all remaining too)

and also - no docstrings had been updated.

ulidtko:
> `# option 3, bad: reintroduce names to global scope...keeps API, but pollutes space`

This isn't bad. Breaking API is bad. Especially so in a dynamically-typed language, without proper deprecation grace periods and major version bumps.

Your error will go away as soon as you reexport the names and stop hiding them inside the enum classes.
https://github.com/urwid/urwid/blob/143ebf41df12f539076ac9a14672f3e8fa6c539d/urwid/container.py#L1784-L1804

penguinolog:
re-implemented in #599

PR #367, opened 2019-10-09
Make Columns and Pile selectable when any child widget is

tu500:
##### Checklist

##### Description:
Currently changing the contents of a `Columns` or a `Pile` widget may break the focus. Specifically the `focus_position` is not updated, making it possible to end up with a focused child widget that is not selectable. This propagates to the Container widget, making it not selectable and therefore not responding to keypresses anymore, thus trapping the focus.
This PR changes the selectability of the 2 container widgets to be selectable whenever _any_ child widget is. An alternative fix would be to need explicit user code for such situations, maybe with something like in #154, but I think this solution is the one with the more expected behaviour.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -82,6 +82,7 @@ Contributors
 - `inducer <//github.com/inducer>`_
 - `winbornejw <//github.com/winbornejw>`_
 - `hootnot <//github.com/hootnot>`_
+- `raek <//github.com/raek>`_
 
 
 .. |pypi| image:: http://img.shields.io/pypi/v/urwid.svg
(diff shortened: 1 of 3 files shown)


tonycpsu:
Thanks for the PR.  Do you have a test program that demonstrates when the focus is trapped?  Pretty sure I've seen the behavior before but can't recall exactly how to recreate for testing.

tu500:
[Quick and dirty example adapting the tour example.](https://github.com/tu500/urwid/commits/columns-selectability)
Focus the middle widget and press F9 vs. focussing the others and pressing F9.

tonycpsu:
Got it.  Fix looks good to me.  Thanks!

PR #372, opened 2019-10-24
Added TrioEventLoop

ntamas:
##### Checklist

##### Description:

This PR adds support for Trio-based event loops. Trio is an `asyncio` alternative for Python 3.5.

The PR implements the same functionality as #298 but with more modern Trio constructs (detached cancel scopes) that were proposed in the comments of #298 by @njsmith

### DIFF
--- a/urwid/_async_kw_event_loop.py
+++ b/urwid/_async_kw_event_loop.py
@@ -112,29 +112,19 @@ class TrioEventLoop(EventLoop):
             True if the scope was cancelled, False if it was cancelled already
             before invoking this function
         """
-        scope, cancelled = scope
-        existed = not cancelled[0]
-        cancelled[0] = True
+        existed = not scope.cancel_called
         scope.cancel()
         return existed
-
-    def _create_cancel_scope(self):
-        """Creates a Trio cancellation scope and a corresponding mutable flag
-        that indicates whether the scope was cancelled already _from_ urwid.
-
-        This is needed because `CancelScope.cancelled_caught` stays `False`
-        until someone actually _handles_ the cancellation, and
-        `CancelScope.cancel_called` can only be called from an async context
-        (which we cannot guarantee).
-        """
-        return self._trio.CancelScope(), [False]
 
     def run(self):
         """Starts the event loop. Exits the loop when any callback raises an
         exception. If ExitMainLoop is raised, exits cleanly.
         """
 
+        idle_callbacks = self._idle_callbacks
+
         def _exception_handler(exc):
+            idle_callbacks.clear()
             if isinstance(exc, ExitMainLoop):
                 return None
             else:
@@ -143,7 +133,7 @@ class TrioEventLoop(EventLoop):
         class TrioIdleCallbackInstrument(self._trio.abc.Instrument):
             def before_io_wait(self, timeout):
                 if timeout > 0:
-                    for idle_callback in self._idle_callbacks.values():
+                    for idle_callback in idle_callbacks.values():
                         idle_callback()
 
         emulate_idle_callbacks = TrioIdleCallbackInstrument()
@@ -220,13 +210,12 @@ class TrioEventLoop(EventLoop):
         Returns:
             a cancellation scope for the Trio task
         """
-        handle = self._create_cancel_scope()
-        scope, _ = handle
+        scope = self._trio.CancelScope()
         if self._nursery:
             self._nursery.start_soon(task, scope, *args)
         else:
             self._pending_tasks.append((task, scope, args))
-        return handle
+        return scope
 
     async def _watch_task(self, scope, fd, callback):
         """Asynchronous task that watches the given file descriptor and calls
--- a/urwid/_async_kw_event_loop.py
+++ b/urwid/_async_kw_event_loop.py
@@ -1,0 +1,244 @@
+#!/usr/bin/python
+#
+# Urwid main loop code using Python-3.5 features (Trio, Curio, etc)
+#    Copyright (C) 2018 Toshio Kuratomi
+#    Copyright (C) 2019 Tamas Nepusz
+#
+#    This library is free software; you can redistribute it and/or
+#    modify it under the terms of the GNU Lesser General Public
+#    License as published by the Free Software Foundation; either
+#    version 2.1 of the License, or (at your option) any later version.
+#
+#    This library is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+#    Lesser General Public License for more details.
+#
+#    You should have received a copy of the GNU Lesser General Public
+#    License along with this library; if not, write to the Free Software
+#    Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+# Urwid web site: http://excess.org/urwid/
+
+from .main_loop import EventLoop, ExitMainLoop
+
+
+class TrioEventLoop(EventLoop):
+    """
+    Event loop based on the ``trio`` module.
+
+    ``trio`` is an async library for Python 3.5 and later.
+    """
+
+    _idle_emulation_delay = 1.0/256 # a short time (in seconds)
+
+    def __init__(self, nursery=None):
+        """Constructor.
+
+        Parameters:
+            nursery: the Trio nursery in which the asynchronous tasks will
+                execute. `None` will make the event loop use `trio.run()` when
+                the loop is started and force it to create a nursery on its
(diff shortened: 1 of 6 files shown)


urwid/_async_kw_event_loop.py
```diff
@@ -0,0 +1,244 @@
...
+        return existed
+
+    def _create_cancel_scope(self):
+        """Creates a Trio cancellation scope and a corresponding mutable flag
+        that indicates whether the scope was cancelled already _from_ urwid.
+
+        This is needed because `CancelScope.cancelled_caught` stays `False` until
+        someone actually _handles_ the cancellation, and
+        `CancelScope.cancel_called` can only be called from an async context
+        (which we cannot guarantee).
```
njsmith: Huh, does urwid use threads? "async context" just means "inside the same thread where `trio.run` is running".

urwid/_async_kw_event_loop.py
```diff
@@ -0,0 +1,244 @@
...
+        """Asynchronous task that sleeps for a short amount of time and then
+        calls all the registered idle callbacks.
+
+        Used to simulate idle callbacks in the Trio event loop that has no
+        concept of being idle.
+        """
+        while True:
+            await self._sleep(self._idle_emulation_delay)
+            for idle_callback in self._idle_callbacks.values():
+                idle_callback()
```
njsmith: It's a bit unfortunate that this needs to wake up constantly, even if there isn't actually any work to do; it hurts battery life b/c it stops CPUs from going to sleep. I think it's fine for a first version, but I do wonder if there's any way to avoid that.

At first I thought well, maybe we can only run the loop while there's an idle function actually registered. But I checked and it looks like urwid always has an idle function registered, so that wouldn't actually help anything.

The reason urwid always has an idle function registered is that it uses it to redraw the screen. Which struck me as a weird thing to do from an idle callback, since normally you'd only want to redraw the screen after something changed, not when the program is sitting there idle!

After looking a bit more I think what's going on here is:

urwid's own home-grown loop calls the idle callbacks *exactly once before each time it sleeps for an extended period*. So the idea is that you go back and forth between executing code, and sleeping; and we assume that the screen might change while executing code, but not while sleeping, so we want to update the screen on those transitions. It's a little bit hacky compared to a more fine-grained system that actually tracks whether the screen needs redraw (e.g. it might do extra redraws, or if the program is busy for a while then it might not redraw for a while), but it's a lot simpler to implement than fine-grained damage tracking, and probably works about 99% as well in practice, so it makes sense.

However, other main loops don't actually provide this operation. E.g. Glib has something called an "idle callback", but it runs constantly whenever the loop could otherwise be sleeping, which is very different. But, since they're both named "idle callback", maybe no-one noticed?

So like... maybe the real solution here is for urwid to implement a fine-grained damage tracking system, which might be non-trivial, but I think it's the only way to really properly support external event loops.

In the mean time...... there *is* a way to get the original run-once-before-sleeping semantics in Trio. It doesn't fix the issue for other loops like glib, and it's pretty gross, but maybe it's a reasonable medium-term hack.

What you could do is use [Trio's instrumentation API](https://trio.readthedocs.io/en/latest/reference-hazmat.html#instrument-api). Normally this is intended for debugging and profiling, but it does let you find out when Trio is [about to go to sleep and run some code](https://trio.readthedocs.io/en/latest/reference-hazmat.html#trio.abc.Instrument.before_io_wait).

Because they're a low-level debugging interface, instrumentation hooks run in a weird context: you have to be very careful not to let any exceptions escape, and you're not inside any task so you can't `await` or use the cancellation machinery normally. (Though you should still be able to do operations that rely on run-global state, like call `trio.current_time()`.)

Anyway, like I said, I don't think you necessarily need to do anything about this now. But it's interesting to think about the options...

ntamas: > It's a bit unfortunate that this needs to wake up constantly, even if there isn't actually any work to do

Yes, I felt bad about this for a while until I realized that `urwid` registers an idle callback for redrawing the screen anyway so there's always at least one idle callback. Other async event loops like the one for Twisted or `asyncio` also handle things in a similar way and I did not know that Trio's event loop actually has a concept similarly to the idle callback of `urwid`'s default `select`-based loop.

I'll investigate whether I could use the instrumentation API for the event loop. In the previous PR, you suggested using `wait_all_tasks_blocked()` instead - wouldn't that be a better option? (At least in the sense that we don't need to register a custom instrument).

njsmith: > In the previous PR, you suggested using wait_all_tasks_blocked() instead - wouldn't that be a better option? (At least in the sense that we don't need to register a custom instrument).

Not really, unfortunately – if you wrote a loop like:

```python
while True:
    await wait_all_tasks_blocked()
    ...
```

...then the loop would indeed wait until the program was otherwise idle before starting to run, *but* once the rest of the program became idle then it would keep running constantly, similarly to the loop you have now.

urwid/_async_kw_event_loop.py
njsmith: > It could be the case that the unit test is doing something fishy (e.g., not actually running the loop but just adding and removing a handle outside the loop), but since this part of the unit test is common for all event loop tests within urwid, it seemed easier not to argue about changing the unit test itself.

I see, yeah, that must be what's going on. If the loop was running then `cancel_called` would work.

It sounds like there's an argument for fixing the test, rather than adding contortions to the main loop to work around these issues, but I can see why you might not want to argue about it too...

urwid/_async_kw_event_loop.py
ntamas: Okay, I have tried to use the instrumentation API (haven't committed it yet), and it seems to work - at least the unit tests pass correctly. Would `before_io_wait()` also get called if there are no IO tasks to wait for at all and there are only pending `sleep()` calls that might wake up the Trio loop?

urwid/_async_kw_event_loop.py
ntamas: > It sounds like there's an argument for fixing the test, rather than adding contortions to the main loop to work around these issues

@tonycpsu @ulidtko Not sure who calls the shots in issues like this; would you be open to changing the unit tests such that the tests for the `remove_alarm()` and `remove_watch_file()` calls (in `test_remove_alarm()` and `test_remove_watch_file()`, respectively) are also executed _while_ the event loop is running? The reason is that right now in the implementation of the Trio event loop we need to jump through extra hoops to make these methods work even _outside_ the event loop. (For me it's okay to keep it this way - after all, it works - I just need a second opinion).

urwid/_async_kw_event_loop.py
njsmith: > Would before_io_wait() also get called if there are no IO tasks to wait for at all and there are only pending sleep() calls that might wake up the Trio loop?

Yeah... the name is maybe a bit confusing, actually. Operating systems provide a single combined check-for-I/O-and-maybe-wait-for-timeouts-to-expire operation, so Trio always uses that. Sometimes Trio just wants to check for I/O without waiting (because there are other tasks already ready to run), sometimes it just wants to wait for timeouts, sometimes it wants to do both at once – in all cases it uses the same underlying operation. And in all cases it calls  `before_io_wait`.

You can tell whether it's going to sleep, actually. If the `timeout` passed in to `before_io_wait` is zero, then it means there are already tasks ready to run, and Trio is just doing a quick check to see if any I/O has appeared – it isn't actually going to wait, despite the name. If the `timeout` argument is > zero, then it means Trio is ready to sleep for a bit.

I guess to really match the traditional urwid semantics, you could make your idle callbacks only run when `timeout > 0`. I'm not sure if this is actually better or not – the tradeoff is that we do fewer screen refreshes, which is more efficient (Trio often does a quick check for I/O even though the program is otherwise busy), but if the program is busy for a while then it could cause the screen refreshes to get out of date.

ntamas: Okay, I have updated my PR with your suggestions - thanks a lot!

urwid/_async_kw_event_loop.py
tonycpsu: @ntamas : I agree with @njsmith that fixing the test code is preferable to complicating the run loop code.  As long as the test code changes don't cause any test failures when testing with the other run loops, I don't have a problem with it.  If it does, then those would need to be addressed, and at that point maybe we just accept what you have that's working now and live to fight another day.

ntamas: Thanks for the feedback - I'll give it a go to refactor the tests and simplify the run loop and I'll push the changes here if it works smoothly.

ntamas: PR updated - I have changed the tests a bit and it now works without additional hackery in the TrioEventLoop code.

Note that I needed to changed the tests for the TwistedEventLoop as well because its reactor cannot be restarted.

PR #384, opened 2019-12-15
Updated documentation on Terminal

### LINKED ISSUE #383: Unicode rendering problems in Terminal
##### Description:
As described in #359, Terminals do not render unicode output correctly. E.g.

```bash
$> echo -en "☺" | xxd
00000000: e298 ba                                  ...
$> echo -e "test \xe2\x98\xba  test"
test ☺  test
```

And then in a Terminal instance:

image

- [x] `master` branch (specify commit)
- [x] Latest stable version from `pypi`
- [ ] Other (specify source)

Have a terminal print unicode glyphs to its stdout. E.g.

```
echo -e "test \xe2\x98\xba  test"
```

##### Expected/actual outcome

The `echo` above should result in this being printed:

```
test ☺  test
```

d0c-s4vage:
##### Checklist

##### Description:

This PR fixes #383 by adding a note in the documentation about setting the global encoding for urwid to utf8 if Terminals aren't rendering unicode glyphs correctly.

### DIFF
--- a/urwid/vterm.py
+++ b/urwid/vterm.py
@@ -1328,29 +1328,36 @@ class Terminal(Widget):
     def __init__(self, command, env=None, main_loop=None, escape_sequence=None,
                  encoding='ascii'):
         """
-        A terminal emulator within a widget.
+        A terminal emulator within a widget. BLAH
 
-        'command' is the command to execute inside the terminal, provided as a
+        ``command`` is the command to execute inside the terminal, provided as a
         list of the command followed by its arguments.  If 'command' is None,
         the command is the current user's shell. You can also provide a callable
         instead of a command, which will be executed in the subprocess.
 
-        'env' can be used to pass custom environment variables. If omitted,
+        ``env`` can be used to pass custom environment variables. If omitted,
         os.environ is used.
 
-        'main_loop' should be provided, because the canvas state machine needs
+        ``main_loop`` should be provided, because the canvas state machine needs
         to act on input from the PTY master device. This object must have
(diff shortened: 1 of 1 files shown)


d0c-s4vage:
The Terminal documentation now looks like this:
image

ulidtko:
First, I need to leave a comment regarding the "backward compatibility" handwaving here in the original doc.

UTF-8 is already backward compatible with ASCII. What exactly would break had this default been changed from `ascii` to `utf8`?

... I don't find any answer to that. Either we'd have "ascii codec can't handle..." crashes no longer happening, or Unicode glyphs rendering correctly by default (by which I mean: in urwid-using programs without `urwid.set_encoding("utf-8")` call). Programs which upgrade from older Urwid would see no new breakage. Programs which used to explicitly call `set_encoding('utf-8')` would see no new breakage.

Perhaps I'm missing something damn obvious; @ntamas, hi! Could you elaborate on what you had in mind with that wording? You'd introduced this pretty recently in PR #349. It just feels like an excuse.

ntamas:
Yes, it's basically an excuse ;) I did not have anything particular in my mind; basically, the old behaviour before `urwid` 2.0.1 was that transmitting non-ASCII characters to the spawned process did not work. I did not know whether `urwid` follows semver conventions or not and I wanted to have this fixed in upstream as soon as possible, so I played it safe to ensure that the PR gets merged into the next minor release at the latest. I'm totally fine with changing the default to `utf-8` (it's the 21st century after all).

PR #385, opened 2019-12-15
Terminal: use UTF-8 by default.

ulidtko:
As I've commented under #384 — there's no such thing as using `ASCII` instead of `UTF-8` for "backward compatibility reasons". That's just plain nonsense.

Having broken Unicode rendering (#383) by default is unacceptable. We should expect *non-utf8* users to discover and call `urwid.set_encoding(our_unfortunate_codepage)` — not the other way around.

This PR fixes that, by changing the default to `utf-8` and removing the unconvincing wording from the docs.

I'll give it a few days to linger. Perhaps the original author @ntamas shows up and explains where I'm being an idiot?..

A few other `vterm`-related editorials are piggybacked as well.

### DIFF
--- a/urwid/vterm.py
+++ b/urwid/vterm.py
@@ -1361,20 +1361,11 @@ class Terminal(Widget):
         """
         self.__super.__init__()
 
-        if escape_sequence is None:
-            self.escape_sequence = "ctrl a"
-        else:
-            self.escape_sequence = escape_sequence
+        self.escape_sequence = escape_sequence or "ctrl a"
 
-        if env is None:
-            self.env = dict(os.environ)
-        else:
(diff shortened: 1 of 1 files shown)


ntamas:
Added a comment in #384 [here](https://github.com/urwid/urwid/pull/384#issuecomment-565840887); basically the bottom line is that you are right, and I'm totally fine with changing the default to UTF-8. (I would have done it the same way from the very beginning had I not been super cautious about not including any modifications that could remotely resemble a breaking change).

ulidtko:
Ok, excellent! Thanks for replying, I totally understand the "play it safe" part :) Cheers

Issue #387, opened 2019-12-16
GridFlow invalid focus handling when v_sep==0


agrenott:
##### Description:
GridFlow keypress propagation is not working properly when setting v_sep to 0.
The inner Pile is incorrectly flagged as NOT selectable (while it should be, as it contains a Column which contains selectable Buttons).
I think this is because the Pile is built empty, then added an empty Column. When the Buttons are added into the Column, it's correctly changed to selectable, but this is not propagated to the outer Pile.
When v_sep is not null, the Pile gets updated by the "del p.contents[:1]" in generate_display_widget, hiding the issue.

##### Affected versions (if applicable)
- 2.1.0

##### Steps to reproduce (if applicable)
```python3
import urwid

def b_pressed(b):
    print(b)

unsure = urwid.Button(u"Unsure", b_pressed)
yes = urwid.Button(u"Yes", b_pressed)
no = urwid.Button( u"No", b_pressed)
# Urwid bug!?!?!
# v_sep=0 breaks button handling
# this makes the Pile used inside GridFlow non-selectable (because content is hidden inside
# Columns, and gets added afterwards).
# Non 0 v_sep works around this issue because of the "del p.contents[:1]" in generate_display_widget
display_widget = urwid.GridFlow([unsure, yes, no], 15, 3, v_sep=0, align='left')
loop = urwid.MainLoop(urwid.Filler(display_widget))
loop.run()
```

##### Expected/actual outcome
b_pressed is never called when v_sep==0, while it should.

PR #388, opened 2019-12-16
Fix GridFlow keypress handling when v_sep is 0

agrenott:
Ensure inner GridFlow's Pile selectable attribute is updated with the
added content, so that keypress events are forwarded properly.

Fixes https://github.com/urwid/urwid/issues/387


ulidtko:
LGTM, merging.

PR #392, opened 2020-01-13
added TrioEventLoop.run_async(), removed nursery constructor arg

ntamas:
##### Checklist

##### Description:

This PR is related to #381 and the discussion therein. When `TrioEventLoop` was introduced, I have mistakenly left a `nursery` constructor argument in the code; the original purpose of this constructor would have been to allow the user to inject an already existing Trio nursery ("task group") into the event loop so all callbacks that are scheduled by `TrioEventLoop` would become scheduled in this nursery instead of a default nursery that `TrioEventLoop` created after calling `run()` on the event loop.

I have never implemented this functionality and forgotten to remove the argument, so the author of #381 complained that the argument does nothing. After some experimentation, I have realized that allowing the user to provide his own nursery would not make sense anyway (because then it would be the user's responsibility to handle `ExitMainLoop` raised by `urwid`, which, when left unhandled, would cancel all other tasks in the nursery as well due to how Trio nurseries work). At the same time, #381 illustrates that there are valid use-cases where an urwid-based terminal UI is embedded in a larger Trio application where Trio is launched already by the time the urwid-based UI is started, so we need to be able to start a `TrioEventLoop` instance _without_ invoking `trio.run()`. Therefore, this PR:

* removes the `nursery` argument of `TrioEventLoop` (which I don't think is a breaking change because it never did anything in the first place)
* adds an asynchronous `run_async()` function to `TrioEventLoop` that runs the event loop in cases when the execution has already entered the async world.

Usage example:

```
async def run_user_interface():
    loop = MainLoop(event_loop=TrioEventLoop(), ...)
    with loop.start():
        await self.loop.event_loop.run_async()

...

async def main():
    with trio.open_nursery() as nursery:
        ...do some other Trio stuff...
        nursery.start_soon(run_user_interface)
        ...do some other Trio stuff...
```

This is the cleanest API I could implement without having to extend the API of `MainLoop` as well with an async `run_async()` function.

### DIFF
--- a/urwid/_async_kw_event_loop.py
+++ b/urwid/_async_kw_event_loop.py
@@ -30,15 +30,8 @@ class TrioEventLoop(EventLoop):
     ``trio`` is an async library for Python 3.5 and later.
     """
 
-    def __init__(self, nursery=None):
-        """Constructor.
-
-        Parameters:
-            nursery: the Trio nursery in which the asynchronous tasks will
-                execute. `None` will make the event loop use `trio.run()` when
-                the loop is started and force it to create a nursery on its
-                own.
-        """
+    def __init__(self):
+        """Constructor."""
         import trio
 
         self._idle_handle = 0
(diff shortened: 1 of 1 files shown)


ulidtko:
LGTM, merging.

PR #414, opened 2020-06-19
Use comparison with a string instead of "is" test with a literal

andrewshadura:
Python 3.8 deprecates `is` test with a literal, which leads to runtime warnings:
```
urwid/numedit.py:245: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if default is not None and default is not "":
```

##### Checklist


tonycpsu:
Thanks, merged.

Issue #377, opened 2019-11-29
unsupported locale setting error in 2.1.0


thornycrackers:
##### Description:
I'm trying to import pudb into my code base with `from pudb import set_trace; set_trace()` and I get this error. 

```
unsupported locale setting error

/opt/venv/lib/python3.6/site-packages/urwid/util.py in detect_encoding
line 58:         locale.setlocale(locale.LC_ALL, initial)
initial | ('en_US', 'UTF-8')
locale | <module 'locale' from '/usr/local/lib/python3.6/locale.py'>


/usr/local/lib/python3.6/locale.py in setlocale
line 598:     return _setlocale(category, locale)
category | 6
locale | 'en_US.UTF-8'
```

If I downgrade to `2.0.1` everything is fine. This is running inside the `python:3.6-slim-stretch` image.
 

##### Affected versions (if applicable)
2.1.0

tonycpsu:
Likely related to #259 , #279 .

hoopes:
I opened a stack overflow for this (before i saw this ticket...): 

https://stackoverflow.com/questions/59295250/python-default-locale-unsupported-locale-setting/59305150#59305150

The locale error from the SO issue was being thrown from urwid when using pudb.

pieqq:
I'm using urwid for a TUI program I'm writing, and I got the same issue when trying to run my program on Ubuntu Core 16.

Here is the output of `locale -a` and `locale` in UC16:

```
$ locale -a
C
C.UTF-8
POSIX

$ locale
locale: Cannot set LC_ALL to default locale: No such file or directory
LANG=C.UTF-8
LANGUAGE=
LC_CTYPE="C.UTF-8"
LC_NUMERIC=en_US.UTF-8
LC_TIME=en_US.UTF-8
LC_COLLATE="C.UTF-8"
LC_MONETARY=en_US.UTF-8
LC_MESSAGES="C.UTF-8"
LC_PAPER=en_US.UTF-8
LC_NAME=en_US.UTF-8
LC_ADDRESS=en_US.UTF-8
LC_TELEPHONE=en_US.UTF-8
LC_MEASUREMENT=en_US.UTF-8
LC_IDENTIFICATION=en_US.UTF-8
LC_ALL=
```

Because of this, when urwid tries to execute the [code that was added for 2.1.0 in util.py](https://github.com/urwid/urwid/pull/279/files#diff-f2004e2753aca67c165b2323b2ef58d4), it fails. This issue can be replicated manually:

[...]

PR #415, opened 2020-06-26
Ignore resetting to invalid locale (Closes: #377)

jspricke:
##### Checklist
~~- [ ] I've included docstrings and/or documentation and/or examples for my code (if this is a new feature)~~

##### Description:
Resetting the locale was introduced in 5dcf236, before that an invalid
locale was ignored. Restore the old behavior insofar that only valid
locales are restored.
This fixes the unit tests of scottkosty/vit with unset LANG.

### DIFF
--- a/urwid/util.py
+++ b/urwid/util.py
@@ -55,7 +55,10 @@ def detect_encoding():
         else:
             raise
     finally:
-        locale.setlocale(locale.LC_ALL, initial)
+        try:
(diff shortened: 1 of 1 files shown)


jspricke:
An example of the broken test can be found here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=963315

Issue #322, opened 2018-11-29
Crash on draw_screen


CameronLonsdale:
##### Description:
In my project https://github.com/CameronLonsdale/MTP I have a list of edit boxes. In one edit box if I spam capital A I eventually get to the following traceback:

```
Traceback (most recent call last):
  File "/Users/clonsdale/Personal/manytime_interactive/venv/bin/mtp", line 11, in <module>
    load_entry_point('mtp', 'console_scripts', 'mtp')()
  File "/Users/clonsdale/Personal/manytime_interactive/cli.py", line 25, in main
    many_time_pad_attack(ciphertexts)
  File "/Users/clonsdale/Personal/manytime_interactive/manytime/__init__.py", line 19, in many_time_pad_attack
    interactive(ciphertexts, partial_key)
  File "/Users/clonsdale/Personal/manytime_interactive/manytime/interactive.py", line 191, in interactive
    Application(ciphertexts, Key(key)).run()
  File "/Users/clonsdale/Personal/manytime_interactive/manytime/interactive.py", line 143, in run
    self._loop.run()
  File "/Users/clonsdale/Personal/manytime_interactive/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 286, in run
    self._run()
  File "/Users/clonsdale/Personal/manytime_interactive/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 384, in _run
    self.event_loop.run()
  File "/Users/clonsdale/Personal/manytime_interactive/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 788, in run
    self._loop()
  File "/Users/clonsdale/Personal/manytime_interactive/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 816, in _loop
    self._entering_idle()
  File "/Users/clonsdale/Personal/manytime_interactive/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 777, in _entering_idle
    callback()
  File "/Users/clonsdale/Personal/manytime_interactive/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 572, in entering_idle
    self.draw_screen()
  File "/Users/clonsdale/Personal/manytime_interactive/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 587, in draw_screen
    self.screen.draw_screen(self.screen_size, canvas)
[...]

PR #416, opened 2020-06-30
Fix bytes/unicode assert error in draw_screen

bugnano:
When running my application that uses Urwid with Python 3 (https://github.com/bugnano/rnr), I ran into this assert error:

  File "/home/fri/.local/lib/python3.8/site-packages/urwid/raw_display.py", line 822, in draw_screen
    assert cs in [None, "0", "U"], repr(cs)
AssertionError: b'0'

By the assert error it was clear that the error was not in my code but in Urwid, as cs is clearly of type bytes, and the assert compares it against string values.

When modifying this code in order to compare cs against byte values instead of strings, my application stopped crashing.

I don't know if this bug has been reported before, so I don't know if this fixes an already reported bug, or not.

### DIFF
--- a/urwid/raw_display.py
+++ b/urwid/raw_display.py
@@ -813,19 +813,19 @@ class Screen(BaseScreen, RealTerminal):
             lasta = lastcs = None
             for (a,cs, run) in row:
                 assert isinstance(run, bytes) # canvases should render with bytes
-                if cs != 'U':
+                if cs != B('U'):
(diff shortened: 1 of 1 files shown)


tonycpsu:
Previously reported as #322, but this fix seems more comprehensive than the one proposed in that issue.

bugnano:
So, any chance of getting this PR merged, and possibly issuing a bugfix release?

PR #418, opened 2020-07-05
Remove idle emulation from asyncio event loop

### LINKED ISSUE #90: High CPU load with AsyncioEventLoop
My project's CPU load is between 8 and 9 % when it's doing absolutely nothing. I've found that setting `AsyncioEventLoop._idle_emulation_delay` from `1/256` to something like `1/20` reduces CPU load to around 1 %, and I don't notice any reduced responsiveness.

Also, `AsyncioEventLoop._we_started_event_loop` isn't used anywhere, so it can be removed?

WGH-:
The main loop documentation for draw_screen() says the following:

> If you modify the widgets displayed outside of handling input or
> responding to an alarm you will need to call this method yourself
> to repaint the screen.

The idle emulation is only used for redrawing screen, so it should
be _correct_ to piggyback on input/alarm callbacks to call idle callbacks
at the end of event loop iteration during which input/alarm callbacks
are called.

Unfortunately, this will break asyncio programs that ignored that
comment and modified urwid state from callbacks registered on asyncio
event loop directly without calling draw_screen(), relying on
entering_idle() being called repeatedly anyway.

A possible fix for them is to use a decorator like the following:

    def call_redraw(f):
        @functools.wraps(f)
        def wrapper(self, *args, **kwargs):
            asyncio.get_event_loop().call_soon(urwid_main_loop.draw_screen)
            f(self, *args, **kwargs)
        return wrapper

It also breaks programs that don't use 'main_loop' argument
of urwid.Terminal constructor despite documentation advising so.

Fixes #90

It's draft because I'm not 100% sure this is the right fix (_and also am lazy to run tox locally_).

### DIFF
--- a/urwid/main_loop.py
+++ b/urwid/main_loop.py
@@ -353,6 +353,11 @@ class MainLoop(object):
         self._reset_input_descriptors()
         self.idle_handle = self.event_loop.enter_idle(self.entering_idle)
 
+        # the screen is redrawn automatically after input and alarms,
+        # however, there can be none of those at the start,
+        # so draw the initial screen here unconditionally
+        self.event_loop.alarm(0, self.entering_idle)
+
         return StoppingContext(self)
 
     def stop(self):
@@ -1380,10 +1385,20 @@ class AsyncioEventLoop(EventLoop):
     ``asyncio`` is new in Python 3.4, but also exists as a backport on PyPI for
     Python 3.3.  The ``trollius`` package is available for older Pythons with
     slightly different syntax, but also works with this loop.
+
+       .. note::
+
+        If you make any changes to the urwid state outside of it
+        handling input or responding to alarms (for example, from asyncio.Task
+        running in background), and wish the screen to be
+        redrawn, you must call :meth:`MainLoop.draw_screen` method of the
+        main loop manually.
+
+        A good way to do this::
+
+            asyncio.get_event_loop().call_soon(main_loop.draw_screen)
     """
     _we_started_event_loop = False
-
-    _idle_emulation_delay = 1.0/30  # a short time (in seconds)
 
     def __init__(self, **kwargs):
         if 'loop' in kwargs:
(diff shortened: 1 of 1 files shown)


WGH-:
> Could you please clarify the motivation of this PR?

The motivation is in #90: the emulation calls the `_entering_idle` function many times per second, which burns the CPU time.

I'll reply to the rest of the questions a bit later.

WGH-:
Sorry for late reply.

> — this isn't entirely true, or maybe I'm reading it in a wrong way. In one urwid-based program I worked on, the .entering_idle callback was used together with timers to implement an "idle logout" feature. 

Okay, perhaps, I worded it in a somewhat ambiguous way (frankly, I don't really understand my point now either). I'll try to explain in cleaner way.

Idle callbacks are called when the event loop returns back to idle state, that is, just after handling all events that have occured.

1. Wait for events (keyboard, timers, etc.)
2. Handle events
3. Call idle callbacks
4. Goto 1

The problem with asyncio event loop is that it doesn't have builtin support of idle callbacks. Right now urwid simply calls them all the time periodically, which is wasteful.

I propose a better emulation of idle callbacks for asyncio: keep track of the events (fd and alarms) registered through urwid event looop, and schedule to call idle callbacks after they occur. To my knowledge, it matches the behaviour of other event loops perfectly, except that user might use asyncio event loop directly, bypassing urwid. Well, the documentation I quoted in the first message already advises against it, so it's likely not really a big deal.

WGH-:
By the way, `Terminal` doesn't work properly on other event loops if you don't pass a `main_loop`. That's kinda expected, and I'd say that fixing it out of scope of this PR.

Still, I do agree that even though it worked without `main_loop` on asyncio event loop just by accident, the fact that it no longer would is a breaking change, and should be mentioned in the changelog.

WGH-:
> Bonus points for stopping to have to rely on the docs instructing how user code should call draw_screen() correctly; this is definite improvement. Could you also clean up the docs?

I don't really understand what can be improved here. The docstring for `draw_screen` is pretty clear in my opinion.

WGH-:
On second thought, I added a short note about that to `AsyncioEventLoop` docstring.

penguinolog:
Re-created with proposed fix and merged

PR #422, opened 2020-07-24
Fix curses_display python3 ord()

amjltc295:
##### Checklist
- [v] I've ensured that similar functionality has not already been implemented
- [v] I've ensured that similar functionality has not earlier been proposed and declined
- [v] I've branched off the `master` or `python-dual-support` branch
- [v] I've merged fresh upstream into my branch recently
- [v] I've ran `tox` successfully in local environment
- [v] I've included docstrings and/or documentation and/or examples for my code (if this is a new feature)

##### Description:
This is actually an issue I faced while using Taurus with oh-my-zsh:
`TypeError: ord() expected string of length 1, but int found`
someone had reported it before:
https://groups.google.com/g/codename-taurus/c/90F2ie0ka0I/m/7iu2mo5yBAAJ
I think this might be a python2 vs python3 issue:
https://stackoverflow.com/questions/50111345/bytes-like-object-is-required-ord-expected-string-of-length-1-but-int-found
https://stackoverflow.com/questions/30076240/ord-expected-string-of-length-1-but-int-found
This change could let Taurus run smoothly with oh-my-zsh

### DIFF
--- a/urwid/curses_display.py
+++ b/urwid/curses_display.py
@@ -32,7 +32,7 @@ from urwid import escape
 
 from urwid.display_common import BaseScreen, RealTerminal, AttrSpec, \
     UNPRINTABLE_TRANS_TABLE
-from urwid.compat import bytes, PYTHON3, text_type, xrange
+from urwid.compat import bytes, PYTHON3, text_type, xrange, ord2
 
 KEY_RESIZE = 410 # curses.KEY_RESIZE (sometimes not defined)
 KEY_MOUSE = 409 # curses.KEY_MOUSE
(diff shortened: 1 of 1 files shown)


urwid/curses_display.py
```diff
@@ -515,8 +515,7 @@ def draw_screen(self, size, r ):
                 try:
                     if cs in ("0", "U"):
                         for i in range(len(seg)):
-                            self.s.addch( 0x400000 +
-                                ord(seg[i]) )
+                            self.s.addch( 0x400000 + seg[i] if PYTHON3 else ord(seg[i]) )
```
WGH-, line 518: `urwid.compat` seems to have `ord2` function defined for that purpose.

amjltc295: Changed to ord2

PR #424, opened 2020-08-09
Use non deprecated template

jspricke:
defindex.html was deprecated a long time ago, see

https://github.com/python/cpython/commit/7970cd483346dfd7723da214fb27399ecc574095

for example.


penguinolog:
Template was removed in sphinx 3, without this changes building of documentation is impossible

Issue #447, opened 2021-01-01
partial native Windows support via windows-curses


asmith-kepler:
### Description
I have managed to get partial support for urwid on Windows 10 and Python 3 using the windows-curses package. A few things don't work, but it works well enough for my needs.

**What works:**
- curses display
- most widgets
- most special keys
- console resizing

**What doesn't work:**
- unicode (including the default symbols for LineBox, etc.)
- mouse events
- high-colour modes (I don't think these are supported by windows-curses)

I don't know enough about the curses library or ANSI control sequences to debug further. I'm posting this here in the hopes that a) someone else finds it useful, and b) someone more knowledgeable can use it as a starting point to fix the remaining issues (especially unicode).

### Patch
```python
"""
Patch urwid on Windows and WSL.

On Windows, install the package "windows-curses".
"""
import curses
import platform
import re

import urwid
import urwid.curses_display
import urwid.raw_display

IS_WSL = ("Linux" in platform.platform()) and ("Microsoft" in platform.platform())
IS_WINDOWS = ("Windows" in platform.platform()) and ("Linux" not in platform.platform())

if IS_WSL:
    # Filter out SI/SO under WSL. See:
    # https://github.com/mitmproxy/mitmproxy/blob/8dfb8a9a7471e00b0f723de66d3b63bd089e9802/mitmproxy/tools/console/window.py#L316-L323
    wsl_patch_orig_write = urwid.raw_display.Screen.write
    wsl_patch_write_re = re.compile("[\x0e\x0f]")

    def wsl_patch_write(self, data):
        data = wsl_patch_write_re.sub("", data)
        return wsl_patch_orig_write(self, data)

    urwid.raw_display.Screen.write = wsl_patch_write

if IS_WINDOWS:
    def win_patch_tty_signal_keys(self, intr=None, quit=None, start=None, stop=None, susp=None, fileno=None):
        pass  # no signals on Windows; not needed for resize
    urwid.display_common.RealTerminal.tty_signal_keys = win_patch_tty_signal_keys

    win_patch_orig__start = urwid.curses_display.Screen._start
    def win_patch__start(self):
        win_patch_orig__start(self)
        # halfdelay() seems unnecessary and causes everything to slow down a lot.
        curses.nocbreak()  # exits halfdelay mode
        # keypad(1) is needed, or we get no special keys (cursor keys, etc.)
        self.s.keypad(1)
    urwid.curses_display.Screen._start = win_patch__start

[...]

PR #448, opened 2021-01-04
Add Windows support for raw_display

mhils:
##### Checklist
- [x] I've ran `tox` successfully in local environment *(yes, but I do not have 2.7, 3.5, and pypy at hand)*

##### Description:

Inspired by @asmith-kepler's fantastic work in #447, I set out to implement Windows support for raw_display. Turns out that this is all working pretty seamlessly with the latest Windows console improvements. 🎉 

Main changes:

 - Use `socket.socketpair()` instead of `os.pipe()` on all OSes as Windows doesn't support non-socket file descriptors in `select()`. This probably has a tiny amount of overhead, but works across all platforms. 
 - Use ctypes to interface with the Windows API, which works without any additional dependencies.
 - Use a dedicated thread to read input from the Windows console and pipe it into another socketpair, which is then processed by raw_display. This is slightly ugly, but works across all events loops.

### DIFF
--- a/urwid/raw_display.py
+++ b/urwid/raw_display.py
@@ -30,13 +30,19 @@ import select
 import struct
 import sys
 import signal
+import socket
+import threading
 
-try:
+
+if os.name == "nt":
+    IS_WINDOWS = True
+    from . import win32
+    from ctypes import byref
+else:
+    IS_WINDOWS = False
     import fcntl
     import termios
     import tty
-except ImportError:
-    pass # windows
 
 from urwid import util
 from urwid import escape
@@ -48,9 +54,11 @@ from urwid.compat import PYTHON3, bytes, B
 
 from subprocess import Popen, PIPE
 
+STDIN = object()
+
 
 class Screen(BaseScreen, RealTerminal):
-    def __init__(self, input=sys.stdin, output=sys.stdout):
+    def __init__(self, input=STDIN, output=sys.stdout):
         """Initialize a screen that directly prints escape codes to an output
         terminal.
         """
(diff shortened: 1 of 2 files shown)


urwid/raw_display.py
```diff
@@ -225,15 +240,35 @@ def _start(self, alternate_buffer=True):
...
 
         fd = self._input_fileno()
-        if fd is not None and os.isatty(fd):
+        if fd is not None and os.isatty(fd) and not IS_WINDOWS:
             self._old_termios_settings = termios.tcgetattr(fd)
             tty.setcbreak(fd)
 
-        self.signal_init()
+        if IS_WINDOWS:
+            hOut = win32.GetStdHandle(win32.STD_OUTPUT_HANDLE)
```
penguinolog, line 248: will be hood if such long branches will be in own methods like:
```python
if IS_WINDOWS:
    self.windows_signal_init()
else:
    self.signal_init()
```

urwid/raw_display.py
```diff
@@ -395,6 +439,10 @@ def unhook_event_loop(self, event_loop):
         """
         Remove any hooks added by hook_event_loop.
         """
+        if self._input_thread is not None:
+            self._input_thread.should_exit = True
+            self._input_thread = None
```
penguinolog, line 444: no `wait` / `join`?

urwid/raw_display.py
```diff
@@ -666,9 +721,18 @@ def get_cols_rows(self):
         y, x = 24, 80
         try:
             if hasattr(self._term_output_file, 'fileno'):
-                buf = fcntl.ioctl(self._term_output_file.fileno(),
-                                termios.TIOCGWINSZ, ' '*4)
-                y, x = struct.unpack('hh', buf)
+                if IS_WINDOWS:
```
penguinolog, line 724: better to make dedicated methods for both branches

penguinolog (changes requested):
please split logic to reduce platform specific `if`/`else` branches size

penguinolog:
Restored updated, tested and merged

mhils:
I totally missed the comments  in March, my apologies. Thanks for taking care of things, this is super duper fantastic! :smiley: :cake: :rocket:

PR #465, opened 2021-02-16
Provide 80x24 fallback for ansi and vt100

roadriverrail:
Not all terminals will necessarily respond to the TIOCGWINSZ ioctl,
particularly much older ones providing vt100 or ansi terminals, which
will have a fixed 80x24 size.  If they respond with some kind of
non-sensical value here, then at least give the ansi and vt100 terminals
their default so they can draw.

##### Checklist
- [*] I've ensured that similar functionality has not already been implemented
- [*] I've branched off the `master` or `python-dual-support` branch
- [*] I've merged fresh upstream into my branch recently
- [*] I've ran `tox` successfully in local environment (my pypy isn't set up, but it's otherwise good)

##### Description:

I've been working with some older machines that present a classic `ansi` or `vt100` terminal, and while `ncurses` worked just fine with them, the `raw_display` from `urwid` did not.  I tracked this down to the use of `TIOCGWINSZ`.  This ioctl wasn't providing meaningful dimensions for these old terminals.  It looked like `urwid` was getting 0x0, while `screen` was getting -1x-1.  These terminal types have a fixed width and height as per their `termcap` descriptions, and they're ubiquitous, so I provided some fallback values should `TIOCGWINSZ` not give workable dimensions.

I tried to see if a patch like this was rejected in the past, but forgive me if I missed it.

### DIFF
--- a/urwid/raw_display.py
+++ b/urwid/raw_display.py
@@ -681,6 +681,10 @@ class Screen(BaseScreen, RealTerminal):
         except OSError:
             # Term size could not be determined
             pass
+        # Provide some lightweight fallbacks in case the TIOCWINSZ doesn't
+        # give sane answers
+        if (x <= 0 or y <= 0) and self.term in ('ansi','vt100'):
+                y, x = 24, 80
         self.maxrow = y
         return x, y
 


urwid/raw_display.py
```diff
@@ -672,6 +672,11 @@ def get_cols_rows(self):
         except IOError:
             # Term size could not be determined
             pass
+        # Provide some lightweight fallbacks in case the TIOCWINSZ doesn't
+        # give sane answers
+        if x <= 0 or y <= 0:
+            if self.term == 'ansi' or self.term == 'vt100':
```
penguinolog, line 678:
```suggestion
            if self.termin('ansi', 'vt100'):
```

penguinolog:
Please rebase: Travis not providing logs (technically not functional) and without logs hard to see what was wrong with current PR

roadriverrail:
Rebased and made the change you suggested (I think).  This was apparently open for like 2 years and I've completely forgotten it existed, so forgive me if I'm a little slow on the uptake here.

urwid/raw_display.py
```diff
@@ -681,6 +681,11 @@ def get_cols_rows(self):
         except OSError:
             # Term size could not be determined
             pass
+        # Provide some lightweight fallbacks in case the TIOCWINSZ doesn't
+        # give sane answers
+        if x <= 0 or y <= 0:
+            if self.term in ('ansi','vt100'):
```
penguinolog, line 687:
```suggestion
        if (x <= 0 or y <= 0) and self.term in ('ansi','vt100'):
```

penguinolog (changes requested):
This 2 it's can be squashed and LGTM

urwid/raw_display.py
roadriverrail: Done.

penguinolog (approved):

PR #473, opened 2021-04-13
Fix column label typo in tour example

devfull:
Minor fix in tour example.


ulidtko (approved):
Nice, good catch! LGTM, tested

PR #494, opened 2022-01-03
pass arguments to SelectableIcon() to parent

doctorcolossus:
`SelectableIcon` subclasses the `Text` widget.
However, it currently isn't possible to pass it `Text` arguments like `wrap`:
`SelectableIcon(text, wrap="ellipsis")`
... because the `SelectableIcon` constructor doesn't pass its arguments along to the `Text` superclass. This commit fixes that.

I'm not sure how or whether to update the docstring, so I haven't changed it. I think it is unwise to simply duplicate all of the possible parameters from the `Text` parent class, because that would violate the DRY principle and introduce the likelihood of their getting out of sync if the `Text` parameters ever change. Feedback welcome.

##### Checklist

##### Description:

### DIFF
--- a/urwid/wimp.py
+++ b/urwid/wimp.py
@@ -36,7 +36,7 @@ from urwid.command_map import ACTIVATE
 class SelectableIcon(Text):
     ignore_focus = False
     _selectable = True
-    def __init__(self, text, cursor_position=0):
+    def __init__(self, text, cursor_position=0, *args, **kwargs):
         """
         :param text: markup for this widget; see :class:`Text` for
                      description of text markup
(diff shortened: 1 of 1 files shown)


urwid/wimp.py
```diff
@@ -36,7 +36,7 @@
 class SelectableIcon(Text):
     ignore_focus = False
     _selectable = True
-    def __init__(self, text, cursor_position=0):
+    def __init__(self, text, cursor_position=0, *args, **kwargs):
```
penguinolog, line 39: Better to accept and provide explicit arguments.
In case of long class hierarchy it will be not comfortable to decompose all `*args` `**kwargs` and lookup how unexpected argument reached base class / where need to manually sanitize.

Extra params for docstring can be directly copied from the docstring of parent class

penguinolog (changes requested):

penguinolog:
re-implemented in #599

PR #524, opened 2023-03-28
Test fixes

penguinolog:
* asyncio coroutine was not awaited, so test was false-positive (poor visible on Travis CI) Fix test.
* FloatEdit unittests passed, incorrect output doctests not caused exception on CI
* warnings on version number
* warnings about deprecated setup.py usage for tests run
* python2.7 is not able to execute tests for asyncio due to mandatory syntax changes (async def). Fix logic: 1. added `isinstance` check before getting significance 2. in case of using points after delimiter, need + 2 to precision 3. properly set precision to context instead of context override 4. normalize version number 5. execute tests via `unittest` 6. stop using python 2.7 in tests run instead of increasing complexity

##### Checklist


urwid/numedit.py
```diff
@@ -247,13 +247,13 @@ def __init__(self, caption="", default=None,
...
-            if isinstance(default, str) and len(default):
+            if isinstance(default, str) and default:
                 # check if it is a float, raises a ValueError otherwise
                 float(default)
                 default = Decimal(default)
 
-            if preserveSignificance:
-                self.significance = abs(default.as_tuple()[2])
+            if preserveSignificance and isinstance(default, Decimal):
+                self.significance = abs(default.as_tuple().exponent) + 1
```
penguinolog, line 256: pre-define known value explicit instead of magic

tox.ini
```diff
@@ -1,30 +1,14 @@
 [tox]
-envlist =
-    py27
-    py35
-    py36
-    py37
-    py38
-    pypy
+envlist = py3{6,7,8,9,10,11},pypy3
```
penguinolog, line 2: use compact form

urwid/tests/test_event_loops.py
```diff
@@ -201,14 +201,14 @@ def test_error(self):
             evl.alarm(0.5, lambda: 1 / 0)  # Simulate error in event loop
             self.assertRaises(ZeroDivisionError, evl.run)
 
-        async def test_coroutine_error(self):
+        def test_coroutine_error(self):
             evl = self.evl
 
             async def error_coro():
```
penguinolog, line 207: this syntax is not supported on python2.
Since sunset of python2, no reason to add magic with compile

urwid/tests/test_event_loops.py
```diff
@@ -201,14 +201,14 @@ def test_error(self):
...
                 result = 1 / 0 # Simulate error in coroutine
-                yield result
```
penguinolog, line 209: yield -> async generator, we need coroutine

urwid/numedit.py
```diff
@@ -265,13 +265,9 @@ def value(self):
...
-            ip, fp = ([v for v in
-                      self.edit_text.split(self._decimalSeparator)] + [0])[0:2]
-
-            with localcontext() as ctx:
-                ctx = (self.significance or len(str(fp)))+1
-                return Decimal(self.edit_text.replace(
-                    self._decimalSeparator, '.')) * 1
+            normalized = Decimal(self.edit_text.replace(self._decimalSeparator, '.'))
+            if self.significance is not None:
+                return normalized.quantize(self.significance)
```
penguinolog, line 270: https://docs.python.org/3/library/decimal.html#decimal.Decimal.quantize should be enough without context magic and manual calculation

penguinolog:
Tests green.
Corner cases now handled correctly, thread context is not modified anymore for Decimal.
Warnings amount reduced (version normalization, not awaited coroutine).

urwid/numedit.py
djyotta: Not to mention I somehow wasn't using the context anyway...
And made the same mistake in the test case (facepalm).

This is a nice clean alternative.

Issue #512, opened 2022-11-10
Comparing str and bytes


dstromberg:
##### Description:
Here's a detailed writeup for pudb, which uses urwid:
https://github.com/inducer/pudb/issues/570

...but in short, running an urwid app with python3 -bb can give:
File "/usr/local/lib/python3.10/site-packages/urwid/display_common.py", line 761, in tty_signal_keys
   if intr == 'undefined': intr = 0
BytesWarning: Comparison between bytes and string

##### Affected versions (if applicable)
$ python3 -m pip list -v | grep -i urwid
below cmd output started 2022 Thu Nov 10 08:16:03 AM PST
urwid              2.1.2       /usr/local/lib/python3.10/site-packages                pip
urwid-readline     0.13        /usr/local/lib/python3.10/site-packages                pip

##### Steps to reproduce (if applicable)
You could likely run urwid's automated tests with python3 -bb.

##### Expected/actual outcome
No traceback.

Issue #408, opened 2020-06-03
Text.set_text() cannot take empty attribute list.


ulidtko:
##### Steps to reproduce
```python
import urwid

t = urwid.Text('')

t.set_text(t.get_text())
```
##### Actual outcome
```python
Traceback (most recent call last):
  File "test.py", line 3, in <module>
    t.set_text(t.get_text())
  File "/home/ulidtko/src/urwid/urwid/widget.py", line 877, in set_text
    self._text, self._attrib = decompose_tagmarkup(markup)
  File "/home/ulidtko/src/urwid/urwid/util.py", line 382, in decompose_tagmarkup
    text = tl[0][:0].join(tl)
IndexError: list index out of range
```

##### Description:
Caused by this hack here: https://github.com/urwid/urwid/blob/f21aee20cc6af1f0ab431b23847d477c3f0359b2/urwid/util.py#L381-L382

This should be done in some different way.

Better yet, just drop Py2-support hacks (#394).

There should also exist a **unit test** for this much trivial bug.

##### Affected versions (if applicable)
- [x] `master` branch (specify commit)
- [x] Latest stable version from `pypi`

Issue #406, opened 2020-04-25
Add type hints (PEP484) and distribute them (PEP561)


1ace:
#### Context

The python ecosystem has several tools to statically analyse your code and (if your editor supports that) pointing out errors like passing the wrong type as you type your code, instead of waiting for a runtime crash.

This analysis relies on developers writing down this information in their code, and having it available in the modules they use. See [PEP 484](https://www.python.org/dev/peps/pep-0484/) for information on those type hints annotations.

#### What about `urwid`?

`urwid` does not provide these hints yet, which means that these tools are unable to check code that uses `urwid`, and instead shows messages like:
```
Skipping analyzing 'urwid': found module but no type hints or library stubs
```

It would be really nice for type annotations ([PEP 484](https://www.python.org/dev/peps/pep-0484/)) to be added to the various files in `urwid`, and distribute them (see [PEP 561](https://www.python.org/dev/peps/pep-0561/)).

Note that this might require #394 (dropping python2 support) to be done first.
I've been fortunate enough to not have to write type hints in legacy code, so I'm not sure how that works; it might be fine, or it might be unusable, or anything in between.

These annotations could for instance be added "as we go" by simply having a policy to not introduce any untyped new code, or maybe a stronger policy of also adding types when touching a bit of existing code. Or someone might be motivated enough to go through entire files at a time and add all the annotations :smile:

#### How do you measure success?

Running `mypy --strict urwid/` in the repository should not return any error (or warning, ideally).

LordAmit:
Looks like urwid will break if type hint is used. 

For example, I wrote code using typehint; people tried to install from PPA; lo and behold - things broke.

```
Setting up python3-urwid (2.0.1-3) ...
/usr/lib/python3/dist-packages/urwid/tests/test_canvas.py:141: SyntaxWarning: 's
tr' object is not callable; perhaps you missed a comma?
  assert result == expected, "got: %r expected: %r" (result, expected)
/usr/lib/python3/dist-packages/urwid/tests/test_canvas.py:145: SyntaxWarning: 's
tr' object is not callable; perhaps you missed a comma?
  assert result == expected, "got: %r expected: %r" (result, expected)
/usr/lib/python3/dist-packages/urwid/tests/test_canvas.py:149: SyntaxWarning: 's
tr' object is not callable; perhaps you missed a comma?
  assert result == expected, "got: %r expected: %r" (result, expected)
[...]

PR #532, opened 2023-04-06
Add extra type annotations

penguinolog:
* add basic mypy config for better tracking (now crazy amount of warnings without `strict`)
* Useless check in `raw_display` (`if not Popen` will be always `False`)
* use explicit `return None`

Partial: #406
Related: #512
Related: #408

##### Checklist

### DIFF
--- a/urwid/vterm.py
+++ b/urwid/vterm.py
@@ -33,6 +33,8 @@ import struct
 import sys
 import time
 import traceback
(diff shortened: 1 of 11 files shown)


urwid/vterm.py
penguinolog, line 1: Another place, where `bytes` compared with `str` :(

PR #537, opened 2023-04-17
[BREAKING CHANGE] Refactoring: Split event loop in several modules

penguinolog:
* `urwid.main_loop` is split into multiple modules which is easier to maintain
* `urwid.compat` is not used anymore and removed
* `TornadoEventLoop`, `GLibEventLoop`, `TwistedEventLoop` and `TrioEventLoop` accessible ONLY if required dependencies installed (like: Tornado installed -> `TornadoEventLoop` is accessible for import)
* `TornadoEventLoop` use the same idle logic as `AsyncioLoop`: tornado.ioloop.IOLoop is asyncio based.
* Trio < 0.15 is not supported. Version 0.15 was released almost 3 years ago.
* Tornado < 5.0 is not supported. Tornado 5.0 was released 5 years ago.

##### Checklist


urwid/event_loop/abstract_loop.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/python
```
wardi, line 1: this line only makes sense when we want a python file to be executable, right?

urwid/event_loop/asyncio_loop.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class AsyncioEventLoop(EventLoop):
+    """
+    Event loop based on the standard library ``asyncio`` module.
+
+    """
+    _we_started_event_loop = False
+
+    _idle_emulation_delay = 1.0/30  # a short time (in seconds)
```
wardi, line 43: it still makes me sad that the asyncio loop needs to consume CPU while idle

wardi: There's an old ticket with a discussion about how this could be fixed if you're interested

urwid/event_loop/abstract_loop.py
penguinolog: Yep. This file was copied as-is, and then splitted.

urwid/event_loop/asyncio_loop.py
penguinolog: I seen this PR, but need to invest extra time to restore PR and address comments.
This will be done as separate PR at short time.

penguinolog:
> What's the breaking change part of this? It looks like the `__init__.py` and conditional imports will remain compatible

Direct import of event loop class from implementation changed.
Stubs import impossible without requirements.
`self.GLib` like hacks removed, probably somebody used it (weird)

urwid/event_loop/abstract_loop.py
```diff
@@ -21,8 +19,11 @@
 #
 # Urwid web site: https://urwid.org/
 
+"""Abstract shared code for urwid IOLoop implementation."""
```
wardi, line 22: IOLoop is a tornado thing, right? I think all the docs call these event loops http://urwid.org/manual/mainloop.html#event-loops

penguinolog: Yep, copy-paste error. Fixed

wardi (approved):
LGTM

PR #541, opened 2023-04-19
[BREAKING CHANGE] Fixes: #90 Remove idle emulation from asyncio event loop

penguinolog:
Re-implement abandoned PR #418
* Fix "not hashable `AttrSpec`" and it's instance creation price (use `__slots__`) `AttrSpec` instances may be created in huge amount, with slots this process consume less resources.
* `Terminal` is always created with event loop, if not provided -> `SelectEventLoop` is used
* Fixed `TornadoEventLoop` & `AsyncioEventLoop` logic (Tornado IOLoop is asyncio based)

For extra details see original PR.

##### Checklist

### DIFF
--- a/urwid/display_common.py
+++ b/urwid/display_common.py
@@ -36,7 +36,7 @@ from urwid import signals
 from urwid.util import StoppingContext, int_scale
 
 if typing.TYPE_CHECKING:
-    from typing_extensions import Literal
+    from typing_extensions import Literal, Self
 
 # for replacing unprintable bytes with '?'
 UNPRINTABLE_TRANS_TABLE = b"?" * 32 + bytes(range(32, 256))
@@ -487,7 +487,7 @@ class AttrSpecError(Exception):
 
 
 class AttrSpec:
-    __slots__ = ("_value",)
+    __slots__ = ("__value",)
 
     def __init__(self, fg: str, bg: str, colors: Literal[1, 16, 88, 256, 16777216] = 256) -> None:
         """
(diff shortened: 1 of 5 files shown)


urwid/display_common.py
```diff
@@ -554,6 +556,13 @@ def __init__(self, fg: str, bg: str, colors: Literal[1, 16, 88, 256, 16777216] =
                 f'require more colors than have been specified ({colors:d}).'
             )
 
+    def __hash__(self) -> int:
+        """Safe hashable.
+
+         Instance is mutable -> use ID component.
+         """
+        return hash((self.__class__, self._value, id(self)))
```
wardi, line 564: if `_value` is mutable should it be used for computing a `__hash__`? Won't this cause a problem if it's used as a key in a dict then modified?

urwid/event_loop/asyncio_loop.py
```diff
@@ -97,38 +137,42 @@ def remove_watch_file(self, handle: int) -> bool:
...
+    def enter_idle(self, callback: Callable[[], typing.Any]) -> int:
         """
         Add a callback for entering idle.
 
-        Returns a handle that may be passed to remove_idle()
+        Returns a handle that may be passed to remove_enter_idle()
         """
         # XXX there's no such thing as "idle" in most event loops; this fakes
         # it the same way as Twisted, by scheduling the callback to be called
         # repeatedly
```
wardi, line 148: update comment?

urwid/event_loop/tornado_loop.py
```diff
@@ -113,36 +138,39 @@ def enter_idle(self, callback: Callable[[], typing.Any]) -> list[object]:
         # XXX there's no such thing as "idle" in most event loops; this fakes
```
wardi, line 138: comment here too

urwid/display_common.py
penguinolog: It's made like done by "unsafe hash" in python data classes. Making instance id specific (not recalculated) hash maybe also solution.
Hash is required for terminal example app.

wardi: interesting, I didn't know about that. Is it worth mentioning this expected unsafe behavior or the reason for adding `__hash__` in the docstring?

wardi: Immutable AttrSpec would be a good change

penguinolog:
> does this change allow handling multiple input/resize events before the screen is redrawn so that the application can "catch up" quickly?

In my test runs it was quick enough during filesystem browse and no issues on widget replacement.

urwid/event_loop/asyncio_loop.py
penguinolog: IMHO it's still correct comment: we schedule callback via `loop.call_later`

wardi:
A test I used to do was steadily drag the terminal window size so the input would fill up with resize events. If the redraw is happening between each event then after you stop resizing the application will take an arbitrarily long time to catch up and be responsive again. If the redraw happens at the end then the application will be almost immediately responsive.

penguinolog:
> A test I used to do was steadily drag the terminal window size so the input would fill up with resize events. If the redraw is happening between each event then after you stop resizing the application will take an arbitrarily long time to catch up and be responsive again. If the redraw happens at the end then the application will be almost immediately responsive.

For me after ~30 seconds of nonstop resize terminal example worked immediately on `AsyncioEventLoop` and `TornadoEventLoop`.

urwid/event_loop/tornado_loop.py
penguinolog: done

urwid/display_common.py
```diff
@@ -545,70 +547,102 @@ def __init__(self, fg: str, bg: str, colors: Literal[1, 16, 88, 256, 16777216] =
...
+
+        if colors is None:
+            new_colors = self.colors
+        else:
+            new_colors = colors
+
+        return self.__class__(foreground, background, new_colors)
+
+    def __hash__(self) -> int:
+        """Instance is immutable and ashable."""
```
wardi, line 583:
```suggestion
        """Instance is immutable and hashable."""
```

wardi (approved):

Issue #542, opened 2023-04-23
Empty row in rendered `Text(..., "center", "clip")` when `maxcol` == `line_width - 1`


AnonymouX47:
##### Description:
When

- `wrap="clip"`,
- `align="center"`,
- there's any line **starting** with text that **has a display attribute** (except via `AttrMap`), and
- the widget's render width (*maxcol*) is one less than the width (in screen columns, not characters) of any of such lines

such a line is rendered as an empty string.

##### Affected versions (if applicable)
- [x] `master` branch (specify commit)
- [x] Latest stable version from `pypi`
- [ ] Other (specify source)

##### Steps to reproduce (if applicable)
```py
w = Text(
    [
        ("attr", "Hello!\nto\nWorld!\n"),
        "Hello!\n",
        ("attr", "to\nWorld!\n"),
    ],
    "center",
    "clip",
)
print([*w.render((5,)).content()])
```

##### Expected/actual outcome
It should render:
```py
[
    [('attr', None, b'Hello')],
    [(None, None, b'  '), ('attr', None, b'to'), (None, None, b' ')],
    [('attr', None, b'World')],
    [(None, None, b'Hello')],
    [(None, None, b'  '), ('attr', None, b'to'), (None, None, b' ')],
    [('attr', None, b'World')],
]
```
but renders:
```py
[
    [],
    [(None, None, b'  '), ('attr', None, b'to'), (None, None, b' ')],
    [],
    [(None, None, b'Hello')],
    [(None, None, b'  '), ('attr', None, b'to'), (None, None, b' ')],
    [],
]
```

##### How it happens
<details>

When the `Text` widget is rendered:
https://github.com/urwid/urwid/blob/350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f/urwid/widget.py#L1061
https://github.com/urwid/urwid/blob/350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f/urwid/widget.py#L1075
https://github.com/urwid/urwid/blob/350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f/urwid/widget.py#L1105
https://github.com/urwid/urwid/blob/350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f/urwid/widget.py#L1114
https://github.com/urwid/urwid/blob/350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f/urwid/text_layout.py#L103

Then the bug originates from:
[...]

PR #543, opened 2023-04-23
Fix text layout for `align="center", wrap="clip"` when `maxcol` == `line_width - 1`

AnonymouX47:
##### Checklist

##### Description:
Fixes: #542

I also realized (and tested) that the bug could be fixed by guarding the append at:
https://github.com/urwid/urwid/blob/350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f/urwid/canvas.py#L1333
with `if s.sc:` but that would only apply to the `Text` widget.

### DIFF
--- a/urwid/tests/test_text_layout.py
+++ b/urwid/tests/test_text_layout.py
@@ -292,6 +292,23 @@ class CalcTranslateClipTest(CalcTranslateTest, unittest.TestCase):
         [(7, None), (0, 35)],
         [(14, 36, 50), (0, 50)]]
 
+class CalcTranslateClipTest2(CalcTranslateTest, unittest.TestCase):
+    text = "Hello!\nto\nWorld!"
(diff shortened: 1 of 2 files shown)


penguinolog:
Please add test to the fixed issue. It should help in case of regression.

AnonymouX47:
Done. Is that sufficient?

AnonymouX47:
If I may ask... Why do so many workflow run attempts always fail?

Just this one PR has spammed my mailbox so dang much... and it has just been two runs so far.

penguinolog:
We have race conditions at test_vterm and event loop test.
Need to fix it.

Op zo 23 apr. 2023 22:50 schreef Toluwaleke Ogundipe <
***@***.***>:

penguinolog (approved):

Issue #544, opened 2023-04-24
Stable tests failure on CI for test_vterm


penguinolog:
Tests normally not fail locally, but stable fail on one or more jobs on CI. Especially in case of parallel run on the same CI machine (normal GitHub scenario).
Looks like resources access race conditions.

penguinolog:
Duplicate #514

PR #545, opened 2023-04-25
Fix incorrect type cast in vterm (`apply_mapping` should return `bytes`)

penguinolog:
Add `time.sleep(0.1)` to the event loop tests:
  in the worst scenario on windows and slow machine
  function in parallel thread/async can wait up to 80 milliseconds (tested)
Add type annotations to the `vterm` and `test_vterm` to simplify error lookup.

Partial: #544
Partial: #512
Partial: #406

##### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,12 @@ universal = 0
 profile = "black"
 line_length = 120
 
+[tool.pytest.ini_options]
+minversion = "6.0"
+addopts = "-vvvv --doctest-modules -s --cov=urwid"
+doctest_optionflags = ["ELLIPSIS", "IGNORE_EXCEPTION_DETAIL"]
+testpaths = ["urwid"]
+
 [tool.coverage.run]
 omit = ["urwid/tests/*", ".tox/*", "setup.py"]
 branch = true
--- a/urwid/tests/test_event_loops.py
+++ b/urwid/tests/test_event_loops.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import asyncio
 import os
 import sys
-import time
 import typing
 import unittest
 
@@ -28,7 +27,6 @@ class EventLoopTestMixin:
         _handle = evl.watch_file(rd, step2)
 
         evl.run()
-        time.sleep(0.1)
         self.assertEqual(out, ["writing", "hi"])
 
     def test_remove_alarm(self):
@@ -89,7 +87,6 @@ class EventLoopTestMixin:
             self.assertEqual(idle_handle, 1)
 
         evl.run()
-        time.sleep(0.1)
         self.assertTrue("waiting" in out, out)
         self.assertTrue("hello" in out, out)
         self.assertTrue("clean exit" in out, out)
@@ -97,7 +94,6 @@ class EventLoopTestMixin:
         del out[:]
 
         evl.run()
-        time.sleep(0.1)
         self.assertEqual(["clean exit"], out)
         self.assertTrue(evl.remove_watch_file(handle))
         _handle = evl.alarm(0, exit_error)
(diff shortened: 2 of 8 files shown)


urwid/wimp.py
```diff
@@ -227,17 +227,17 @@ def set_state(self, state: bool | Literal['mixed'], do_callback: bool = True) ->
         do_callback -- False to suppress signal from this change
 
         >>> changes = []
-        >>> def callback_a(cb, state, user_data):
+        >>> def callback_a(user_data, cb, state):
```
penguinolog, line 230: @wardi looks like incorrect resolve order in `Signals.emit`: `cb` is part of `*args` -> put after `user_args` and before deprecated `user_arg`.

pyproject.toml
```diff
@@ -49,6 +49,12 @@ universal = 0
 profile = "black"
 line_length = 120
 
+[tool.pytest.ini_options]
```
penguinolog, line 52: now preparation only: warnings collected much better, but non-stable event loop  tests and fully incompatible `test_vterm`

urwid/widget.py
```diff
@@ -801,12 +801,13 @@ def render(self, size: tuple[int], focus: bool = False):
         return canv
 
 
-class SolidFill(BoxWidget):
```
penguinolog, line 804: many years deprecated base class

urwid/vterm.py
```diff
@@ -202,25 +202,25 @@ def activate(self, g: int) -> None:
...
         self._sgr_mapping = False
         self.activate(g=self.active)
 
-    def apply_mapping(self, char):
+    def apply_mapping(self, char: bytes) -> bytes:
         if self._sgr_mapping or self._g[self.active] == 'ibmpc':
             dec_pos = DEC_SPECIAL_CHARS.find(char.decode('cp437'))
             if dec_pos >= 0:
                 self.current = '0'
-                return str(ALT_DEC_SPECIAL_CHARS[dec_pos])
```
penguinolog, line 223: incorrect cast, will cause `BytesWarning`

urwid/tests/test_vterm.py
```diff
@@ -226,6 +240,7 @@ def test_erase_display(self):
         self.expect('   x5a98765')
 
     def test_scrolling_region_simple(self):
+        # TODO(Aleksei): Issue #544
```
penguinolog, line 243: 4 tests which fail on regular basis

urwid/tests/test_event_loops.py
```diff
@@ -11,38 +13,48 @@
...
             out.append(os.read(rd, 2).decode('ascii'))
             raise urwid.ExitMainLoop
-        handle = evl.alarm(0, step1)
-        handle = evl.watch_file(rd, step2)
+
+        _handle = evl.alarm(0, step1)
+        _handle = evl.watch_file(rd, step2)
+
         evl.run()
+        time.sleep(0.1)
```
wardi, line 31: good to add explanation of why this is needed as a comment here

urwid/tests/test_vterm.py
wardi: we can mark them as flaky or to be skipped so they don't fail unrelated changes, can't we?

urwid/vterm.py
```diff
@@ -1376,10 +1375,10 @@ def __init__(
...
         A terminal emulator within a widget.
 
-        ``command`` is the command to execute inside the terminal, provided as a
-        list of the command followed by its arguments.  If 'command' is None,
-        the command is the current user's shell. You can also provide a callable
-        instead of a command, which will be executed in the subprocess.
+        ``command`` is the command to execute inside the terminal,
+        provided as a list of the command followed by its arguments.
+        If 'command' is None, the command is the current user's shell.
+        You can also provide a callable instead of a command, which will be executed in the subprocess.
```
wardi, line 1381: seems like you *almost* aligned this paragraph to a new line width

urwid/tests/test_event_loops.py
penguinolog: Extra validation shown ineffectiveness of sleep

urwid/vterm.py
penguinolog: I changed split to increase readability: misread caused incomplete type annotation

wardi (approved):

PR #546, opened 2023-04-26
Skip flaky tests until rewrite

penguinolog:
Tests failing with 90% probability is seriously annoying and block development. Issue should receive priority for proper fix,
until fix tests should be executed separately

##### Checklist

Related #544

### DIFF
--- a/urwid/tests/test_vterm.py
+++ b/urwid/tests/test_vterm.py
@@ -238,6 +238,7 @@ class TermTest(unittest.TestCase):
         self.write(r'98765\e[8D\e[1Jx')
         self.expect('   x5a98765')
 
+    @unittest.skip("Issue #544")
     def test_scrolling_region_simple(self):
         # TODO(Aleksei): Issue #544
         self.write('\\e[10;20r\\e[10f1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\\e[faa')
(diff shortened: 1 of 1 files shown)


penguinolog:
> These don't look flaky to me.
> 
> The test results are pretty consistent in CI, 3 of 3 pipelines I checked failed in the same way:
> 
> image
> 
> Judging by the truncated output from `self.expect()` — there might be an issue with testing library; but that won't explain the ordering of `aa` and `\n\n\n...` getting flipped from expected.

This four tests fail consistently the same way on CI, not fail on local `tox` run and pass on n-th retry on CI.
Need detail investigation for source of this behavior

penguinolog:
@ulidtko run with incremental tests (for extra verbosity): https://github.com/penguinolog/urwid/actions/runs/4834531715
and locally (exactly pushed branch):
```
Ran 345 tests in 0.521s

OK
```
Behavior looks like some data just not transferred or corrupted:
```
<esc>[10;20r -> csi_set_scroll(10, 20)
<esc>[10f -> move_cursor(0, 9, False, False, False)
```
Should set initial state for text, but scroll region set incorrectly/cursor set incorrectly.
Probably rewrite test with mocking of `sys.stdout` will help, but it's not 5 minutes change

penguinolog:
Found solution for incomplete data

PR #547, opened 2023-04-30
Optimize `vterm`: adopt data types and add annotations. Fix tests

### LINKED ISSUE #544: Stable tests failure on CI for test_vterm

penguinolog:
* `TermModes` - make `dataclass` as it's used
* `TermScroller` - deprecate and use `deque` instead
  (faster both side access, automatic size limit control)
* `utf8_buffer` - use `bytearray` as mutual storage for bytes
  (no re-allocation if no crazy size changes)
* `CSI_COMMANDS` - use `namedtuple` types for better type management
* use native string escape sequences instead of hex encoded one
  (easier to read and understand, internally the same)
* not `elif` after `return` (follow actual static checkers rules)

* `_target_encoding = 'ascii'` as default (no re-cast)
* add extra verbose version of tests for scrolling region

Fix #544 

##### Checklist

### DIFF
--- a/urwid/util.py
+++ b/urwid/util.py
@@ -74,7 +74,7 @@ if 'detected_encoding' not in locals():
 else:
     assert 0, "It worked!"
 
-_target_encoding = None
+_target_encoding = 'ascii'
 _use_dec_special = True
 
 
--- a/urwid/vterm.py
+++ b/urwid/vterm.py
@@ -34,32 +34,35 @@ import sys
 import time
 import traceback
 import typing
-from collections.abc import Iterable, Mapping, Sequence, Callable
+import warnings
+from collections import deque
+from collections.abc import Callable, Iterable, Mapping, Sequence
+from dataclasses import dataclass
 
 try:
     import fcntl
     import pty
     import termios
 except ImportError:
-    pass # windows
+    pass  # windows
 
-from urwid import util
+from urwid import event_loop, util
 from urwid.canvas import Canvas
 from urwid.display_common import _BASIC_COLORS, AttrSpec, RealTerminal
 from urwid.escape import ALT_DEC_SPECIAL_CHARS, DEC_SPECIAL_CHARS
 from urwid.widget import BOX, Widget
-from urwid import event_loop
 
 if typing.TYPE_CHECKING:
     from typing_extensions import Literal
 
 EOF = b''
 ESC = chr(27)
+ESC_B = b'\x1b'
 
 KEY_TRANSLATIONS = {
-    'enter':     chr(13),
+    'enter':     "\r",
     'backspace': chr(127),
-    'tab':       chr(9),
+    'tab':       "\t",
     'esc':       ESC,
     'up':        f"{ESC}[A",
     'down':      f"{ESC}[B",
@@ -98,7 +101,19 @@ KEY_TRANSLATIONS_DECCKM = {
     'f5':        f"{ESC}[15~",
 }
 
-CSI_COMMANDS = {
+
+class CSIAlias(typing.NamedTuple):
+    alias_mark: str  # can not have constructor with default first and non-default second arg
+    alias: bytes
+
+
+class CSICommand(typing.NamedTuple):
+    num_args: int
+    default: int
+    callback: Callable[[TermCanvas, list[int], bool], typing.Any]  # return value ignored
+
+
+CSI_COMMANDS: dict[bytes, CSIAlias | CSICommand] = {
     # possible values:
     #     None -> ignore sequence
     #     (<minimum number of args>, <fallback if no argument>, callback)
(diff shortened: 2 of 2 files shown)


urwid/tests/test_vterm.py
```diff
@@ -238,6 +243,58 @@ def test_erase_display(self):
         self.write(r'98765\e[8D\e[1Jx')
         self.expect('   x5a98765')
 
+    def test_scrolling_region_simple_incremental(self):
```
penguinolog, line 246: this extra verbose tests maybe removed or replace "short" versions.
@wardi which approach is better?

urwid/util.py
```diff
@@ -74,7 +74,7 @@ def detect_encoding() -> str:
 else:
     assert 0, "It worked!"
 
-_target_encoding = None
+_target_encoding = 'ascii'
```
penguinolog, line 77: it's set implicit, make explicit

urwid/vterm.py
```diff
@@ -34,32 +34,35 @@
...
 if typing.TYPE_CHECKING:
     from typing_extensions import Literal
 
 EOF = b''
 ESC = chr(27)
+ESC_B = b'\x1b'
 
 KEY_TRANSLATIONS = {
-    'enter':     chr(13),
+    'enter':     "\r",
```
penguinolog, line 63: more human-readable with the same internals

urwid/vterm.py
```diff
@@ -98,7 +101,19 @@
     'f5':        f"{ESC}[15~",
 }
 
-CSI_COMMANDS = {
+
+class CSIAlias(typing.NamedTuple):
```
penguinolog, line 105: named tuples less complex to deal with

urwid/vterm.py
```diff
@@ -260,20 +300,37 @@ def __init__(self, width: int, height: int, widget: Terminal) -> None:
...
         self.has_focus = False
 
-        self.scrollback_buffer = TermScroller()
+        self.scrollback_buffer: deque[list[tuple[AttrSpec | None, str | None, bytes]]] = deque(maxlen=10000)
         self.scrolling_up = 0
 
-        self.utf8_eat_bytes = None
-        self.utf8_buffer = b''
+        self.utf8_eat_bytes: int | None = None
+        self.utf8_buffer = bytearray()
```
penguinolog, line 310: `bytearray` changed in-place
`bytes` is immutable and new object on any change

urwid/vterm.py
```diff
@@ -1690,16 +1755,14 @@ def keypress(self, size: tuple[int, int], key: str) -> str | None:
...
-                key = KEY_TRANSLATIONS_DECCKM.get(key)
+                key = KEY_TRANSLATIONS_DECCKM[key]
             else:
                 key = KEY_TRANSLATIONS.get(key, key)
 
         # ENTER transmits both a carriage return and linefeed in LF/NL mode.
-        if self.term_modes.lfnl and key == "\x0d":
-            key += "\x0a"
-
-        key = key.encode(self.encoding, 'ignore')
```
penguinolog, line 1701: do not change type of variable just for 1 line, make in-place during call

urwid/tests/test_vterm.py
```diff
@@ -101,6 +104,7 @@ def resize(self, width: int, height: int, soft: bool = False) -> None:
     def write(self, data: str) -> None:
         data = data.encode('iso8859-1')
         self.command.write(data.replace(br'\e', b'\x1b'))
+        sleep(0.05)
```
penguinolog, line 107: looks like after write to `stdout` buffer, it's not enough time for full processing and read

urwid/tests/test_vterm.py
```diff
@@ -140,7 +145,7 @@ def expect(
         else:
             desc += '\n'
         desc += f'Expected:\n{what!r}\nGot:\n{got!r}'
-        self.assertEqual(got, what, desc)
+        self.assertEqual(what, got, desc)
```
penguinolog, line 148: Most unittest parsers expect, that expected value is first

urwid/vterm.py
```diff
@@ -1690,16 +1755,14 @@ def keypress(self, size: tuple[int, int], key: str) -> str | None:
                 key = chr(ord(key[-1]) - ord('A') + 1)
         else:
             if self.term_modes.keys_decckm and key in KEY_TRANSLATIONS_DECCKM:
-                key = KEY_TRANSLATIONS_DECCKM.get(key)
+                key = KEY_TRANSLATIONS_DECCKM[key]
```
penguinolog, line 1758: we already know that it in `KEY_TRANSLATIONS_DECCKM`

urwid/tests/test_vterm.py
```diff
@@ -60,6 +63,7 @@ def read(self, size: int) -> bytes:
 
     def write(self, data: bytes) -> None:
         os.write(self.writer, data)
+        sleep(0.025)
```
ulidtko, line 66: Not fond of this. Never really fixes flaky tests — merely shifts flake probability.

Did you try `os.fdopen(self.writer).flush()` instead ?

Or better yet, disabling buffering completely with `unbuffered_pipe = os.fdopen(self.writer, 'wb', 0)

ulidtko (changes requested):

urwid/tests/test_vterm.py
penguinolog: We can try unbuffered pipe, but `stdout` will be still buffered

ulidtko (commented):
The naming of added tests is incomprehensible.

* `test_scrolling_region_simple_incremental`
* `test_cursor_scrolling_region_incremental`
* `test_scrolling_region_simple_with_focus_incremental`
* `test_cursor_scrolling_region_with_focus_incremental`

When seen side-by-side — one can sort-of-make-out-maybe-guesstimate how they differ.

When seen pages apart — no chance :confused:

urwid/tests/test_vterm.py
penguinolog: `os.fdopen` fails in 10 from 10 runs with `Bad file descriptor`
`os.fsync` also disagree to deal with pipe

penguinolog:
> The naming of added tests is incomprehensible.
> 
> * `test_scrolling_region_simple_incremental`
> * `test_cursor_scrolling_region_incremental`
> * `test_scrolling_region_simple_with_focus_incremental`
> * `test_cursor_scrolling_region_with_focus_incremental`
> 
> When seen side-by-side — one can sort-of-make-out-maybe-guesstimate how they differ.
> 
> When seen pages apart — no chance :confused: 

Tests was added only for debug purposes and now it's possible to decide: keep or drop

urwid/tests/test_vterm.py
ulidtko: That can be disabled too.

In a unit test though, why is there a dependency onto `sys.stdout` at all? Can't it allocate a private PTY ?

... Even that seems redundant, as `.read()` passes the output of `Terminal.render()` to asserts directly:

https://github.com/urwid/urwid/blob/bac20db87eaeb0b7a0026807d7e6dbe391b08580/urwid/tests/test_vterm.py#L118-L129

I don't see any need to use `sys.stdout` at all.

penguinolog:
Without huge mocking, `Terminal.spawn`  call `pty.fork`, which accept 0
arguments.

Op ma 1 mei 2023 20:14 schreef Maxim Ivanov ***@***.***>:

urwid/tests/test_vterm.py
penguinolog: Render spawn process, which execute `pty.fork()`.
Data read from fork stdout and rendered.

ulidtko:
> `Terminal.spawn`  call `pty.fork`, which accept 0 arguments.

Yes, that's the `Terminal`'s internal pty. This is fine and expected.

What I'm talking about is `DummyCommand` in test_vterm touching `sys.stdout`. It a dummy "command" running in the terminal and doing IO to its PTY just for tests, why does it need sys.stdout?

penguinolog:
> > `Terminal.spawn`  call `pty.fork`, which accept 0 arguments.
> 
> Yes, that's the `Terminal`'s internal pty. This is fine and expected.
> 
> What I'm talking about is `DummyCommand` in test_vterm touching `sys.stdout`. It a dummy "command" running in the terminal and doing IO to its PTY just for tests, why does it need sys.stdout?

`DummyCommand` `sys.stdout` output points to PTY, not to system real stdout

wardi (approved):

Issue #548, opened 2023-04-30
Current default branch breaks khal


dvzrv:
##### Description:
Building from latest commit on default branch (master) breaks khal.
As discussed on https://github.com/urwid/urwid/commit/9976f338c122a208b8a9108590ea525086cdd5a1 I am currently trying to find a commit of this project which works on Python 3.11 and does not break the consumers of this package (see downstream issue with Arch Linux: https://bugs.archlinux.org/task/78333).

##### Affected versions (if applicable)
`master` branch https://github.com/urwid/urwid/commit/1cc1d63358f7759785b39079eca069ed41078b6d (broken since at least https://github.com/urwid/urwid/commit/350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f)

##### Steps to reproduce (if applicable)
```sh
$ ikhal
Traceback (most recent call last):
  File "/usr/bin/ikhal", line 5, in <module>
    main_ikhal()
  File "/usr/lib/python3.11/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/khal/cli.py", line 501, in interactive_cli
    controllers.interactive(
  File "/usr/lib/python3.11/site-packages/khal/controllers.py", line 591, in interactive
    pane = ui.ClassicView(
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/khal/ui/__init__.py", line 1065, in __init__
    calendar = CalendarWidget(
[...]

PR #549, opened 2023-05-01
Return original code to the deprecated getters and setters

penguinolog:
* Make more aggressive deprecation warning

Fix: #548

##### Checklist

### DIFF
--- a/urwid/listbox.py
+++ b/urwid/listbox.py
@@ -133,9 +133,9 @@ class SimpleListWalker(MonitoredList, ListWalker):
             f"Method `{self.__class__.__name__}._get_contents` is deprecated, "
             f"please use property`{self.__class__.__name__}.contents`",
             DeprecationWarning,
-            stacklevel=2,
+            stacklevel=3,
         )
-        return self.contents
+        return self
 
     def _modified(self):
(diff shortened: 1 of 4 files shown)


urwid/tests/test_vterm.py
```diff
@@ -60,6 +61,7 @@ def read(self, size: int) -> bytes:
 
     def write(self, data: bytes) -> None:
         os.write(self.writer, data)
+        sleep(0.025)
```
penguinolog, line 64: fix for stable failing tests, btw it's also present in dedicated to vterm PR

urwid/listbox.py
```diff
@@ -133,9 +133,9 @@ def _get_contents(self):
             f"Method `{self.__class__.__name__}._get_contents` is deprecated, "
             f"please use property`{self.__class__.__name__}.contents`",
             DeprecationWarning,
-            stacklevel=2,
+            stacklevel=3,
         )
-        return self.contents
+        return self
```
penguinolog, line 138: == `self.contents`, but have to maintain copy-paste during deprecation period

wardi (approved):

Issue #554, opened 2023-05-07
Error when using BigText: urwid.canvas.CanvasError: Attribute extends beyond text


andresmrm:
##### Description:
After upgrading from Python 3.10 to 3.11, I had to reinstall urwid, from 2.1.2 to **2.1.3**.
Now using BigText, even from the example, gives the error below.
I tested other examples (graph, tour, fib) and they work.

I see now that the latest stable version is 2.1.2. Not sure why ArchLinux python-urwid package installed 2.1.3.
Installing 2.1.2 using pip solved the issue for me, even with Python 3.11.

```
$ python bigtext.py
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/urwid/font.py", line 243, in render
    canv = TextCanvas(byte_lines, None, character_set_lines, maxcol=width, check_width=False)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/urwid/canvas.py", line 402, in __init__
    raise CanvasError(f"Attribute extends beyond text \n{text[i]!r}\n{attr[i]!r}" )
urwid.canvas.CanvasError: Attribute extends beyond text 
b'      '
[(None, 16)]

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/tmp/a/bigtext.py", line 163, in <module>
    main()
  File "/tmp/a/bigtext.py", line 160, in main
    BigTextDisplay().main()
  File "/tmp/a/bigtext.py", line 144, in main
    self.loop.run()
  File "/usr/lib/python3.11/site-packages/urwid/event_loop/main_loop.py", line 320, in run
    self._run()
  File "/usr/lib/python3.11/site-packages/urwid/event_loop/main_loop.py", line 423, in _run
    self.event_loop.run()
  File "/usr/lib/python3.11/site-packages/urwid/event_loop/select_loop.py", line 148, in run
    self._loop()
  File "/usr/lib/python3.11/site-packages/urwid/event_loop/select_loop.py", line 185, in _loop
    alarm_callback()
  File "/usr/lib/python3.11/site-packages/urwid/event_loop/main_loop.py", line 621, in entering_idle
    self.draw_screen()
  File "/usr/lib/python3.11/site-packages/urwid/event_loop/main_loop.py", line 635, in draw_screen
[...]

PR #555, opened 2023-05-08
Fix `TextCanvas` `CanvasError("Attribute extends beyond text...")`

penguinolog:
* `[[]] * ...` causes list of 1 list with pointers amount equal to multiplier instead of "list of lists"
* Add 2 basic font tests which check for Canvas create issue
* Add few type annotations during debug process
* Force tests to restore default encoding in tearDown

Fix: #554

##### Checklist

### DIFF
--- a/urwid/tests/test_font.py
+++ b/urwid/tests/test_font.py
@@ -6,6 +6,13 @@ import urwid
 
 
 class TestFontRender(unittest.TestCase):
+    def setUp(self) -> None:
+        self.old_encoding = urwid.util._target_encoding
+        urwid.set_encoding("utf-8")
+
+    def tearDown(self) -> None:
+        urwid.set_encoding(self.old_encoding)
+
     def test_001_basic(self):
         font = urwid.Thin3x3Font()
         rendered = b'\n'.join(font.render("1").text).decode()
--- a/urwid/tests/test_text_layout.py
+++ b/urwid/tests/test_text_layout.py
@@ -1,9 +1,22 @@
 from __future__ import annotations
 
+import contextlib
 import unittest
+from collections.abc import Generator
 
 import urwid
 from urwid import text_layout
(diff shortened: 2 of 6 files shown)


urwid/curses_display.py
```diff
@@ -597,10 +597,10 @@ class FakeRender: pass
         attr += [[],[],[]]
         cols,rows = self.ui.get_cols_rows()
         keys = None
-        while keys!=['q']:
-            r.text=([t.ljust(cols) for t in text]+[""]*rows)[:rows]
-            r.attr=(attr+[[]]*rows) [:rows]
```
penguinolog, line 602: place for the same issue

urwid/tests/test_font.py
```diff
@@ -0,0 +1,30 @@
...
+    def tearDown(self) -> None:
+        urwid.set_encoding(self.old_encoding)
+
+    def test_001_basic(self):
+        font = urwid.Thin3x3Font()
+        rendered = b'\n'.join(font.render("1").text).decode()
+        expected = ' ┐ \n │ \n ┴ '
+        self.assertEqual(expected, rendered)
+
+    def test_002_non_rect(self):
```
penguinolog, line 22: this test should crash without fix for issue

urwid/tests/test_text_layout.py
```diff
@@ -33,8 +46,12 @@ class CalcBreaksCharTest(CalcBreaksTest, unittest.TestCase):
 
 class CalcBreaksDBCharTest(CalcBreaksTest, unittest.TestCase):
     def setUp(self):
+        self.old_encoding = urwid.util._target_encoding
         urwid.set_encoding("euc-jp")
```
penguinolog, line 50: this encoding leaked to the other tests without cleanup stage

PR #558, opened 2023-06-01
Added highcolor support to vterm (88,256,truecolor.) Fixes #457

### LINKED ISSUE #457: urwid.Terminal widget terminal emulator not supporting >16 colors correctly
##### Description:
When I use the urwid.Terminal widget, only the 16 ANSI colors are working correctly, 88-color, 256-color and 24-bit color results in blinking monochrome characters instead.

After taking a look at the code in question, I see that it was mostly written back in 2010 and 2011, so I assume that the lack of support for >16 color terminal emulation is probably because it has not been implemented, rather than because of a bug.

If anyone is up for implementing support for "modern" terminal emulation, including 88/256/24-bit color (when running within a terminal that has 88/256/24-bit color support, of course), I would be happy to pay a bounty (within reason) to that developer / those developers. It would be a great feature to have for me personally, and I'm sure there are others that would appreciate this feature as well.

Latest version from github (master branch, latest commit 58e8278d8a36ad5ecba1e22af63d8490a42477c7)

The easiest way to reproduce it is to run examples/palette_test.py from within examples/terminal.py.

##### Expected/actual outcome
Expected outcome: Correct color output, no garbled characters etc
Actual outcome: Monochrome blinking characters and other weirdness

danschwarz:
##### Checklist
- [X] I've ran `tox` successfully in local environment  [PYTHON 3.X ONLY; can't run 2.7]

##### Description:
Adds high color support to vterm.py - 88 color, 256 color, true color modes.
Monochrome and 16 color mode work as before. 

What's missing/needed:
Testing against Python 2.7 - my Ubuntu 22.04 dev box complains when running 2.7, which causes tox to fail.
Stress testing, especially against pathological cases - bad ANSI escapes.
Add tox tests *if* it's possible to verify correct color output using tox tests?

Here is palette_test.py running in terminal.py:
image

Note, high color and 256 color text will only display properly if you are running on a terminal that is able to support that color mode, and is configured to do so. Old versions of tmux, screen may cause trouble. Be sure your terminal properly displays the color mode by running palette_test.py outside of terminal.py before running it within terminal.py.


danschwarz:
closing; will resubmit with a clean PR against latest master

PR #564, opened 2023-06-06
Build system: use SCM for version info

penguinolog:
Use `setuptools-scm` package for version build
For docs build need to enable "install package" on RTD `python -m build` and `pip install` works without changes

##### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -447,3 +447,6 @@ GitHub.sublime-settings
 
 ### GPG template
 secring.*
+
+# Generated version
+/urwid/version.py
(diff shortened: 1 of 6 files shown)


wardi:
#567 says we have a regression here. How are our tags affecting the version that setuptools-scm generates?

penguinolog:
Setuptools_scm generate version:
If tag set (canonical version) - use tag
If tag not set -
<previous_tag>.dev<commits_from_previous_tag>.<info_about_last_commit_date>

Regression was not from build system change, but at dependency storage
change at `Canvas`. Commit with regression was reverted as standard
solution for mission-critical project

Op do 8 jun. 2023 03:57 schreef Ian Ward ***@***.***>:

PR #576, opened 2023-06-16
Adds support for bracketed paste mode, fixes #452. Also, adds bracketed paste passthru for vterm.

### LINKED ISSUE #452: Bracketed paste mode support
##### Description:

When pasting multi-line clipboard contents into urwid, the newline symbols (`\n`) get reported as `enter` key presses. 

Most terminal emulators now support the [bracketed paste mode](https://cirw.in/blog/bracketed-paste), which should allow urwid to distinguish between the Enter key presses and newlines in pasted text.

##### Affected versions 
- [x] `master` branch (`4c739b6be21b0c98324e7b7780d8e712f3ad6db3`)
- [x] Latest stable version from `pypi` (`2.1.2`)

##### Steps to reproduce 

```python
import urwid
def unhandled_input(key):
    txt.set_text(' '.join([txt.text, repr(key)]))
txt = urwid.Text('pressed keys:')
fill = urwid.Filler(txt)
loop = urwid.MainLoop(fill, unhandled_input=unhandled_input)
loop.run()
```

(same idea: can use the Global Input [example](https://urwid.readthedocs.io/en/latest/tutorial/index.html#global-input) in urwid tutorial, or run [`input_test.py`](https://github.com/urwid/urwid/blob/release-2.1.2/examples/input_test.py) from `examples`)
 
Pasting (with e.g. `Shift + Insert`) into the terminal window where urwid is running a multi-line clipboard contents:

```
Ab
Cd
```

The output is
```
pressed keys: 'A' 'b' 'enter' 'C' 'd' 'enter'
```

##### Expected/actual outcome

Expected: pasted newline symbols are interpreted as `\n` "keys".
Actual: pasted newline symbols are interpreted as `enter` keys.

##### Related

Support bracketed paste mode escape sequences #119

danschwarz:
##### Checklist

##### Description:

Note that bracketed paste mode is _NOT automatically enabled_ for applications; this would cause problems for any apps that aren't bracketed-paste-aware.  

There's an example of how to enable it in your urwid application in examples/terminal.py.

Then you need to actually handle the 'paste begin' / 'paste end' key sequences in your application's keypress method.  There's no example code for this yet. 

For more info on bracketed paste mode, see Wikipedia
or  https://cirw.in/blog/bracketed-paste

### DIFF
--- a/examples/twisted_serve_ssh.py
+++ b/examples/twisted_serve_ssh.py
@@ -46,7 +46,7 @@ from twisted.conch.manhole_ssh import (
 )
 from twisted.cred.portal import Portal
 from twisted.python.components import Adapter, Componentized
-from zope.interface import Attribute, Interface, implements
+from zope.interface import Attribute, Interface, implementer
 
 import urwid
 from urwid.raw_display import Screen
@@ -84,8 +84,6 @@ class IUrwidMind(Interface):
         """Refresh the UI"""
 
 
-
-
 class UrwidUi:
 
     def __init__(self, urwid_mind):
@@ -112,7 +110,6 @@ class UrwidUi:
         return loop
 
 
-
 class UnhandledKeyHandler:
 
     def __init__(self, mind):
@@ -132,9 +129,8 @@ class UnhandledKeyHandler:
         self.mind.terminal.loseConnection()
 
 
+@implementer(IUrwidMind)
 class UrwidMind(Adapter):
-
-    implements(IUrwidMind)
 
     cred_checkers = []
     ui = None
@@ -158,9 +154,6 @@ class UrwidMind(Adapter):
 
     def draw(self):
         self.ui.loop.draw_screen()
-
-
-
 
 
 class TwistedScreen(Screen):
@@ -189,8 +182,7 @@ class TwistedScreen(Screen):
         self._pal_escape = {}
         self.bright_is_bold = True
         self.register_palette_entry(None, 'black', 'white')
-        urwid.signals.connect_signal(self, urwid.UPDATE_PALETTE_ENTRY,
-            self._on_update_palette_entry)
+        urwid.signals.connect_signal(self, urwid.UPDATE_PALETTE_ENTRY, self._on_update_palette_entry)
         # Don't need to wait for anything to start
         self._started = True
 
@@ -368,14 +360,14 @@ class UrwidTerminalProtocol(TerminalProtocol):
         """
         self.urwid_mind.push(data)
 
-    def _unhandled_input(self, input):
+    def _unhandled_input(self, data):
         # evil
         proceed = True
         if hasattr(self.urwid_toplevel, 'app'):
-            proceed = self.urwid_toplevel.app.unhandled_input(self, input)
+            proceed = self.urwid_toplevel.app.unhandled_input(self, data)
         if not proceed:
             return
-        if input == 'ctrl c':
+        if data == 'ctrl c':
             self.terminal.loseConnection()
 
 
@@ -390,7 +382,7 @@ class UrwidUser(TerminalUser):
     The default implementation doesn't
     """
     def __init__(self, original, avatarId):
-        TerminalUser.__init__(self, original, avatarId)
+        super().__init__(original, avatarId)
         self.avatarId = avatarId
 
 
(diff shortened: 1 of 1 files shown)


danschwarz:
Given that the only part of this PR that actually lives in the core urwid libraries are the constants:

```
ENABLE_BRACKETED_PASTE_MODE = f"{ESC}[?2004h"
DISABLE_BRACKETED_PASTE_MODE = f"{ESC}[?2004l"
```
should I just move these constants to `examples\terminal.py`?  If I do this, the code sample (and apps written similarly) will be compatible with a wider range of urwid versions.

penguinolog:
IMHO constants should be added to the main urwid tree (as added in PR to the `vterm` + `escape`
Subclass for screen is questionable (I'll check later, but probably extend Screen API by adding optional constructor flag should be enough).

danschwarz:
Ok I'll await your decision. Thinking about adding an example of handling the 'begin paste', 'end paste' sequences in examples/edit.py app.

examples/terminal.py
```diff
@@ -23,6 +23,26 @@
 
 import urwid
 
+from urwid import raw_display, escape
+
+class BracketedPasteScreen(urwid.raw_display.Screen):
```
penguinolog, line 28: I re-checked `Screen` code and IMHO we can extend it for bracketed paste support directly

danschwarz:
OK, updated the PR to add an optional bracketed_paste_mode parameter:
```
class Screen(BaseScreen, RealTerminal):
    def __init__(self, input=sys.stdin, output=sys.stdout, bracketed_paste_mode=False):
```

and I've updated the `examples\terminal.py` to use the parameter rather than subclassing Screen.

examples/edit.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+w#!/usr/bin/env python
```
penguinolog, line 1: looks like unplanned change

danschwarz: Yikes. I'll fix that!

penguinolog (approved):

wardi:
Is there ever a reason to have bracketed paste mode disabled?

danschwarz:
> Is there ever a reason to have bracketed paste mode disabled?

Yes. Bracketed paste mode is designed as a per-app opt-in feature (by Thomas Dickey and collaborators) because if bracketed paste mode is enabled for an app that doesn't have code to explicitly handle the paste start/end markers  `{ESC}[200~` and `{ESC}[201~`, those markers show up in the app's text field along with the pasted text.  

Or, in the case of urwid,  they are translated to `begin paste` and `end paste` strings, and those strings will surround the pasted text.

Bracketed paste mode is a terminal-wide setting, so a well-behaved application will enable bracketed paste mode at startup and disable it at shutdown.

In responding to your comment, I started to think more carefully about my bracketed-paste passthru code in vterm.py.  The vterm terminal behavior does not change in response to `{ESC}[?2004h`  or `{ESC}[?2004l` .  It always passes through any `begin paste` and `end paste` strings from the host terminal.  This is not ideal because it does not allow an application running in vterm to opt-in to bracketed paste mode.  

I will add an additional change in `vterm.py` to this PR so that `{ESC}[200~` and `{ESC}[201~` are *not* passed through unless the application running in vterm.py explicitly opts in through use of `{ESC}[?2004h`. These sequences will be 'swallowed'.  If an application opts in, the sequences will be passed through.  I think this is the right approach.

danschwarz:
By the way, I thought of fadding some kind of basic handling to the Edit widget.  I think the most useful thing that could be done is to automatically highlight the pasted text, as is done in zsh which supports bracketed paste mode. The text would get un-highlighted when the user types anything into the Edit widget before, after, or within the pasted text section.  Unfortunately the Edit widget highlighting behavior seems to be disabled or only partially implemented?  See: https://github.com/urwid/urwid/blob/aa2e14fefc353261420922b1a63e5ffcdb77f981/urwid/widget.py#L1709

I don't know that code very well so I'm reluctant to mess with it.

penguinolog:
@danschwarz can you add dedicated test for this functionality, please?
(in `test_vterm`)

danschwarz:
I am not sure it is possible to add an automated test for this functionality as it is entirely dependent on enabling and using a feature of the host terminal (the feature may or may not exist, depending on the host terminal emulator in use, and the only way that use it is by executing manual cut/paste commands against the host terminal app)

I'll think about whether it's possible to simulate parts of the functionality in an automated way. 🤔

danschwarz:
Oy. I'm not sure what I did with `git` to force close this PR and remove all code.  Probably best to open up a new PR.  This one will include automated tests for bracketed paste mode @penguinolog

PR #580, opened 2023-06-20
CD: Change code to use trusted publishing

penguinolog:
* PyPI project changes is mandatory to support trusted publishing

https://docs.pypi.org/trusted-publishers/adding-a-publisher/


wardi (approved):
Have to trust you on this one, never used this feature before

PR #595, opened 2023-08-17
Split widget and introduce base enums

penguinolog:
Split base widget (`widget` module), container widgets (`container` module) and decoration widgets (`decoration` and `graphics` modules) into several modules under subpackage `widget`.

Introduce string-compatible enums for **sizing**, **align**, **valign**, **wrap mode** and base **width-height settings**.
Enforce automating imports sorting and code formatting + static validation for introduced changes.

##### Checklist


urwid/widget/constants.py
penguinolog, line 1: technically after enums introduction and move of sizing helpers it's not constants, but I have no idea about better naming

urwid/widget/wimp.py
penguinolog, line 1: I was lazy and not splitted this module: it's < 1000 LOC's and can be splitted later if will be too complex

urwid/listbox.py
penguinolog, line 1: automatically re-formatted after imports change (no manual work)

urwid/tests/test_signals.py
```diff
@@ -25,6 +26,7 @@ def test_connect(self):
         edit.set_edit_text('another text')
         handler.assert_not_called()
 
+    @unittest.skipIf(sys.implementation.name == "pypy", "WeakRef works differently on PyPy")
```
penguinolog, line 29: this test is expecially unstable on local PyPy run.
Also manual experiments shown, that everything garbage collector related work can work differently on PyPy

urwid/tests/test_doctests.py
```diff
@@ -9,9 +9,30 @@
 
 def load_tests(loader, tests, ignore):
     module_doctests = [
-        urwid.widget,
-        urwid.wimp,
-        urwid.decoration,
+        urwid.widget.attr_map,
```
penguinolog, line 12: doctests need explicit paths :(

penguinolog:
Validated compatibility against downstream packages: khal and zulip-terminal

urwid/widget/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+from .box_adapter import BoxAdapter, BoxAdapterError
+from .columns import Columns, ColumnsError
+from .constants import (
+    RELATIVE_100,
+    Align,
+    Sizing,
+    VAlign,
+    WHSettings,
+    WrapMode,
+    normalize_align,
```
penguinolog, line 16: normalize_* / simplify_* are useful in downstream code, but was never exported "public"

urwid/widget/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+    WidgetWrapError,
+    delegate_to_widget_mixin,
+    fixed_size,
+    nocache_widget_render,
+    nocache_widget_render_instance,
+)
+from .widget_decoration import WidgetDecoration, WidgetDisable, WidgetPlaceholder
+from .wimp import Button, CheckBox, CheckBoxError, RadioButton, SelectableIcon
+
+__all__ = (
```
penguinolog, line 56: `__all__` needed due to backwards-compatibility constants definition (as it's not used locally, it's defined as "public" explicit)

JC3:
"clip" no longer works for the width parameter to `Padding` because of this commit (specifically, because of the conversion to enums).

```
  File "/Users/cipriani_pro2023/pyenv-brixel-newprotobuf/lib/python3.9/site-packages/urwid/widget/padding.py", line 137, in __init__
    self._width_type, self._width_amount = normalize_width(width, PaddingError)
  File "/Users/cipriani_pro2023/pyenv-brixel-newprotobuf/lib/python3.9/site-packages/urwid/widget/constants.py", line 269, in normalize_width
    raise err(
urwid.widget.padding.PaddingError: width value 'clip' is not one offixed number of columns, 'pack', ('relative', percentage of total width), 'clip'
```

The root cause is that `WHSettings.CLIP` is assigned a value of `WrapMode.CLIP`, which causes the `in` statement at https://github.com/penguinolog/urwid/blob/9e1a128bf80805ee84d8cb1af5b5b5017817240d/urwid/widget/constants.py#L259 to always evaluate to `false`.

Note that the following example program will print `False`:

```
import enum

class OtherTest(str, enum.Enum):
    CLIP = "clip"
    
class Test(str, enum.Enum):
    PACK = "pack"
    CLIP = OtherTest.CLIP
    
print("clip" in (Test.PACK, Test.CLIP))
```

The fix is to assign `WHSettings.CLIP` to the actual string value `"clip"`.

I feel this PR really ought to have been more thoroughly tested before being merged.

The consequence at the moment is that release 2.2.0 can't be used in programs tat use `"clip"` for `Padding`.

ethanfurman:
Using the `StrEnum` in Python 3.12 does function as expected.

penguinolog:
On python 3.10 synthetic tests it also worked, but release is for python 3.7+ - need to fix now and wait for old python releases EOL.

ethanfurman:
Further explanation: When a type is mixed in to an enum, that type is used to generate the `.value`, so the value of `Test.CLIP` was generated by `str(OtherTest.CLIP)` (which is `"OtherTest.CLIP"`).  In Python 3.12 `StrEnum` was added, and `IntEnum` and `IntFlag` modified, so that `str(enum_member)` was the same as `str(enum_member.value)` precisely to avoid the problem of constants converted to enums no longer working correctly you are seeing now.

Having said that, using one enum member as the value of another enum member is not intended, and can easily cause confusion:

```python
import enum

class OtherTest(enum.StrEnum):
    CLIP = "clip"

class Test(enum.StrEnum):
    PACK = "pack"
    CLIP = OtherTest.CLIP

"clip" in (Test.PACK, Test.CLIP))
# True
Test.CLIP == OtherTest.CLIP
# True
Test.CLIP is OtherTest.CLIP  # given the above definition, one could expect this to be True
# False
```

Issue #421, opened 2020-07-21
Status on pack method implementation for basic widget classes


preetmishra:
Currently, among other basic widgets classes, only the `Text` widget class has its `pack` method implemented.

While the documentation has a note regarding the implementation, it would be worthwhile to know about its current priority for the project.

#### [Note](http://urwid.org/reference/widget.html#urwid.Widget.pack)
> This is a new method that hasn’t been fully implemented across the standard widget types. In particular, it has not yet been implemented for container widgets.

PR #621, opened 2023-09-25
Feature: Support pack() for CheckBox/RadioButton/Button

penguinolog:
Partial #421

##### Checklist


wardi (approved):
This is ok, but maybe add a comment in the constructor that the pack method below assumes the wrapped widget won't change.

Possibly better would be to implement pack in Columns so that we don't have to define it here, don't need special warnings in the code and make it useful in more cases.

Issue #603, opened 2023-09-12
NumEdit doesn't allow negative numbers...


JC3:
##### Description:

NumEdit does not allow negative numbers for input. It does not accept the `-` sign.

Problem is here https://github.com/urwid/urwid/blob/f5a9b71da810ef27018244ae434a18ad7d5a9fe5/urwid/numedit.py#L41 and in subsequent use of `ALLOWED`.

Example:

```
import urwid
import urwid.numedit

txt = urwid.numedit.FloatEdit()
fill = urwid.Filler(txt, 'top')
loop = urwid.MainLoop(fill)
loop.run()
```

Steps:
1. Run program
2. Attempt to enter "-1"

Expected: "-1" entered

Actual: "-" is missed, only "1" is entered.

##### Affected versions (if applicable)

2.1.2

Issue #432, opened 2020-09-14
Negative number is not allowed in urwid.IntEdit


chaogebaba:
Can't input '-' in IntEdit object.

Issue #502, opened 2022-03-09
FloatEdit doesn't handle floats and IntegerEdit returns Decimals


oliver-s-lee:
##### Description:
Two issues I've run into regarding numedit widgets that I wanted to ask about. I've rolled into one because I suspect they have a common underlying basis:

1. The FloatEdit widget doesn't accept floats as a default value and its `value()` method returns Decimals, not sure if this is intended or not but at the very least it confused me considering the name of the widget.
2. The IntegerEdit widget's `value()` method also returns Decimals, this one feels more like a bug to me but unsure if also intended.

##### Affected versions (if applicable)
- [ ] `master` branch (specify commit)
- [x] Latest stable version from `pypi`
- [ ] Other (specify source)

##### Steps to reproduce (if applicable)
```
>>> import urwid.numedit
>>> urwid.numedit.IntegerEdit(default = 1).value()
Decimal('1')
>>> urwid.numedit.FloatEdit(default = 1.1)
ValueError: default: Only 'str', 'int', 'long' or Decimal input allowed
```
##### Expected/actual outcome
```
>>> import urwid.numedit
>>> urwid.numedit.IntegerEdit(default = 1).value()
1
>>> urwid.numedit.FloatEdit(default = 1.1)
```

PR #636, opened 2023-09-26
Refactoring numedit: PEP8 arguments, allow negative, type casts

penguinolog:
* USE PEP8 compliant arguments and deprecate old one
* Allow cast `IntEdit` to int and `FloatEdit` to float
* Allow negative values without changing default behavior

Fix: #603
Fix: #432
Partial: #502

##### Checklist

### DIFF
--- a/urwid/numedit.py
+++ b/urwid/numedit.py
@@ -22,10 +22,14 @@
 from __future__ import annotations
 
 import re
-from collections.abc import Container
+import warnings
 from decimal import Decimal
+from typing import TYPE_CHECKING
 
 from urwid import Edit
+
+if TYPE_CHECKING:
+    from collections.abc import Container
 
 
 class NumEdit(Edit):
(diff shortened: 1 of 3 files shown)


urwid/numedit.py
```diff
@@ -196,14 +253,18 @@ def __init__(
...
         caption="",
         default: str | int | Decimal | None = None,
-        preserveSignificance: bool = True,
-        decimalSeparator: str = '.',
+        preserveSignificance: bool | None = None,
+        decimalSeparator: str | None = None,
+        *,
+        preserve_significance: bool = True,
+        decimal_separator: str = ".",
+        allow_negative: bool = False,
```
JC3, line 261: This appears to be a breaking change to code that used to use `preserveSignificance` and `decimalSeparator` as a keyword arg. The following changes should probably be made to preserve compatibility:

- Keep `preserveSignificance` defaulted to `True`
- Have `preserve_significance` default to `None`
- Assign `preserve_significance = preserveSignificance` if `preserve_significance == None`

And the same for `decimalSeparator`.

Otherwise, from the code, it looks like the value of `preserveSignificance` is ignored, which is the breaking part.

Either that or just leave it be and accept that the case doesn't fit the normal style.

Issue #649, opened 2023-09-29
[docs] docs/tutorial/attr.py break tutorial load if trio installed


penguinolog:
##### Description:
```
python docs/tutorial/adventure.py
Traceback (most recent call last):
  File "/mnt/docs/tutorial/adventure.py", line 3, in <module>
    import urwid
  File "/mnt/urwid/__init__.py", line 76, in <module>
    from urwid.event_loop import AsyncioEventLoop, EventLoop, ExitMainLoop, MainLoop, SelectEventLoop
  File "/mnt/urwid/event_loop/__init__.py", line 34, in <module>
    from .trio_loop import TrioEventLoop
  File "/mnt/urwid/event_loop/trio_loop.py", line 30, in <module>
    import trio
  File "/usr/local/lib/python3.12/site-packages/trio/__init__.py", line 19, in <module>
    from ._core import TASK_STATUS_IGNORED as TASK_STATUS_IGNORED  # isort: skip
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/trio/_core/__init__.py", line 9, in <module>
    from ._entry_queue import TrioToken
  File "/usr/local/lib/python3.12/site-packages/trio/_core/_entry_queue.py", line 4, in <module>
    import attr
  File "/mnt/docs/tutorial/attr.py", line 15, in <module>
    txt = urwid.Text(('banner', " Hello World "), align='center')
          ^^^^^^^^^^
AttributeError: partially initialized module 'urwid' has no attribute 'Text' (most likely due to a circular import)
```

In case of `attr.py` rename, all examples loaded correctly

Version 2.1.2 is not affected.
In case of no `trio` package installed, tutorial works without issues

Details: `trio` import `attr`  from stdlib, but in case of tutorial usage it's directory is automatically added to `sys.path` as first entry.
This results instead of python stdlib `attr` is `attr` from tutorial is being imported (stdlib shadowed).

##### Affected versions (if applicable)
- [X] `master` branch (specify commit)
- [x] Latest stable version from `pypi`

PR #653, opened 2023-09-29
Fix tutorial: rename `attr` to `urwid_attr`

penguinolog:
Prevent stdlib `attr` shadowing in case of `sys.path` altered

Fix: #649

##### Checklist


penguinolog:
@mwhudson without this PR changes, full regeneration of images in documentation is impossible :-(

wardi (approved):
interesting, is there something special about the name attr?

penguinolog:
> interesting, is there something special about the name attr?

I mixed in my memory dataclasses with "attrs" package.
`attr` is a part of `attrs` package, it is extremely common library for cases, where dataclasses are not available or not enough functional.
You define like 
```python

import attr

@attr.s
class Config:
    a: dict[str, int] = attr.ib(factory=dict, ...
```
And other code is generated

wardi:
so this change isn't actually required then

penguinolog:
> so this change isn't actually required then

Scripts location is added to `sys.path` at first position, then during imports modules lookup uses `sys.path`. Anything using `attr` is broken and it's one of our dependencies, which call some logic on import

wardi:
the docs/tutorial directory isn't added to sys.path and these are scripts not modules to be imported, right?

penguinolog:
> the docs/tutorial directory isn't added to sys.path and these are scripts not modules to be imported, right?

It's added automatically on run as script.
Short example
```
alekseis@1857be7-lcelt ~/tmp $ batcat *          
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: logging.py
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ print("this should not be executed")
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: tst.py
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ import logging
   2   │ import sys
   3   │ 
   4   │ print(sys.path)
   5   │ 
   6   │ logging.basicConfig(level=logging.INFO)
   7   │ logger = logging.getLogger(__name__)
   8   │ logger.info("it works")
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
alekseis@1857be7-lcelt ~/tmp $ python tst.py     
this should not be executed
['/home/alekseis/tmp', '/usr/lib/python311.zip', '/usr/lib/python3.11', '/usr/lib/python3.11/lib-dynload', '/usr/local/lib/python3.11/dist-packages', '/usr/lib/python3/dist-packages', '/usr/lib/python3.11/dist-packages']
  File "/home/alekseis/tmp/tst.py", line 6, in <module>
    logging.basicConfig(level=logging.INFO)
    ^^^^^^^^^^^^^^^^^^^
AttributeError: module 'logging' has no attribute 'basicConfig'
```
As you can see, `sys.path` was altered and instead of normal logging, another code was imported

PR #655, opened 2023-10-02
Refactor: join display modules in package

penguinolog:
* Simplify support, when amount of code will raise
* `RealTerminal.tty_signal_keys`: make windows-specific branch
* Do not deprecate: `curses_display` and `raw_display` imports

##### Checklist

Partial #447


urwid/lcd_display.py
penguinolog, line 1: Special case, can be prepared for deprecation (no deprecation at this moment)

urwid/html_fragment.py
penguinolog, line 1: Special case, can be prepared for deprecation (no deprecation at this moment)

urwid/escape.py
penguinolog, line 1: Was used internally, old location deprecated

urwid/display_common.py
penguinolog, line 1: Base class and helpers, directly used internally.
Main public code is accessed from package namespace without changes.

urwid/__init__.py
```diff
@@ -194,10 +194,16 @@
...
 )
 
-from . import event_loop, widget
+from . import display, event_loop, widget
 
 from urwid.treetools import ParentNode, TreeListBox, TreeNode, TreeWalker, TreeWidget, TreeWidgetError  # isort: skip
 
+# Optional displays with external dependencies
+try:
+    from .display import curses as curses_display  # backward compatibility
```
penguinolog, line 203: this way we have full backward compatibility without extra files.
Even `PendingDeprecation` is too much for such common module at this moment

wardi:
-1 from me. This is a small library, I don't think we need more namespaces. Also things like `urwid.display.curses` just will shadow the core `curses` module, right?

urwid/display/web.py
```diff
@@ -594,8 +119,9 @@
 ]
 
 
-class Screen:
+class Screen(BaseScreen):
```
penguinolog, line 122: Subclass base class to warranty interfaces

urwid/display/web.py
```diff
@@ -722,17 +256,6 @@ def stop(self):
     def set_input_timeouts(self, *args):
         pass
 
-    def run_wrapper(self, fn):
```
penguinolog, line 725: already in base class implemented correct way

penguinolog:
> -1 from me. This is a small library, I don't think we need more namespaces.
After "clean-up" we have a bit too many files for root module.

> Also things like `urwid.display.curses` just will shadow the core `curses` module, right?
No, if nothing forces `urwid.display` as part of `sys.path` (like using it as root directory for script call or manual `sys.path` modifications), `absolute_import` will not allow to shadow.

urwid/display/_web.js
penguinolog, line 1: After extract from python code, static analysis complains about deprecated APIs usage

urwid/display/_web.css
```diff
@@ -0,0 +1,12 @@
+body {    margin: 8px 8px 8px 8px; border: 0;
+    color: black; background-color: silver;
+    font-family: fixed; overflow: hidden; }
```
penguinolog, line 3: I have warning "Font property font-family does not have generic default" here

urwid/__init__.py
```diff
@@ -242,3 +244,46 @@
...
+# Backward compatible lazy load with deprecation warnings
+_moved_warn: dict[str, str] = {
+    "escape": "urwid.display.escape",
+    "lcd_display": "urwid.display.lcd",
+    "html_fragment": "urwid.display.html_fragment",
+    "web_display": "urwid.display.web",
+}
+# Backward compatible lazy load without any warnings
+# Before DeprecationWarning need to start PendingDeprecationWarning process.
+_moved_no_warn: dict[str, str] = {
```
penguinolog, line 261: IMHO, no any warnings/deprecation should be before the next API breaking release (when `DeprecationWarning` will be dropped and `PendingDeprecationWarning` became `DeprecationWarning`).
Then should be long cycle of deprecation (`PendingDeprecationWarning` -> `DeprecationWarning`).
This is too wide used API

urwid/display/web.py
```diff
@@ -624,7 +150,15 @@ def register_palette(self, palette):
                 raise KeyError(f"palette entry '{like_name}' doesn't exist")
             self.palette[name] = self.palette[like_name]
 
-    def register_palette_entry(self, name, foreground, background, mono=None):
+    def register_palette_entry(
```
penguinolog, line 153: do not break API

urwid/display/web.py
penguinolog, line 1: `str.join` for HTML page generation looks like great candidate for templating usage

penguinolog:
Tested for compatibility with success:
pudb
nomadnet
socli
TextWorld
s-tui
mitmproxy

examples/lcd_cf635.py
```diff
@@ -312,7 +312,7 @@ def build_submenu(ms):
     return build_submenu(menu_structure)
 
 
-screen = urwid.lcd_display.CF635Screen(sys.argv[1])
+screen = urwid.display.lcd_display.CF635Screen(sys.argv[1])
```
wardi, line 315: this is `display.lcd` now, isn't it?

wardi (approved):

Issue #264, opened 2018-01-01
urwid WSL alignment and symbols


foobarbecue:
I'm delighted that urwid works fairly well on the default Ubuntu bash terminal of the windows subsystem for linux. However, there seems to be some sort of charset issue:

```
import urwid

buttons = [urwid.Button(str(msg)) for msg in range(5)]
right = urwid.Filler(urwid.Pile([urwid.AttrMap(button, None,'dark') for button in buttons]))
loop = urwid.MainLoop(right, palette=[('light','white','black'),
                                        ('dark','black','white')])
loop.run()
```
![image](https://user-images.githubusercontent.com/854789/34472042-a4c8c664-ef0c-11e7-91b1-fdf6ad79e289.png)

Does anybody know what's going on here?

mhils:
FYI, we had similar (the same?) issue for @mitmproxy and fixed it like this: https://github.com/mitmproxy/mitmproxy/blob/8dfb8a9a7471e00b0f723de66d3b63bd089e9802/mitmproxy/tools/console/window.py#L316-L323. 

Side note: It's fantastic to see that urwid seems to get some maintenance love again. Thanks, @and3rson!

and3rson:
@mhils Thanks for the info! Can you please also provide the following data:
- Value of `os.getenv('TERM')` (or `echo $TERM`) within the terminal
- Does this reproduce on other windows bash shells? (i. e. Git Bash)
Thanks!

mhils:
`$TERM` is `xterm-256color` with the default WSL shell, 
`platform.platform()` is `Linux-4.4.0-43-Microsoft-x86_64-with-Ubuntu-16.04-xenial`. 
I don't have a comprehensive overview over other shells (the default one is good enough to be frank), but I just downloaded Cmder, which seems to have the same issue:

![image](https://user-images.githubusercontent.com/1019198/35099881-ec7b7682-fc59-11e7-9b10-ff2236c09e2a.png)

Same $TERM and platform here.

foobarbecue:
Not sure if you still wanted someone to check, but the problem does not happen in git bash on windows -- display is fine and `echo $TERM` returns `xterm` (instead of the `xterm-256color` that I get with default WSL shell). Note that in order to get this to work in git bash I have to run the python shell on another machine and ssh to it, because I can't get urwid to work on git bash on windows.

and3rson:
[...]

PR #656, opened 2023-10-02
Fix WSL support: filter-out SI/SO in case of WSL

penguinolog:
##### Checklist

Fix: #264


penguinolog:
> IIRC SO/SI are used for some graphics characters when not in UTF-8 mode:
> https://github.com/urwid/urwid/blob/671ba131114cf86e79ebf76cd10d01c553e6054a/urwid/escape.py#L47C1-L47C55
> 
> instead of stripping these control characters at the last minute it's probably better to disable all the graphics character handling before sending it to the screen (graphics all get converted to `?` or some other placeholder on WSL)

WSL is in UTF8. I validated on real windows.
All characters is displayed correctly.

penguinolog:
rebased from latest `master`

penguinolog:
Validated that changes are not needed for curses display under WSL

PR #664, opened 2023-10-23
Internal: minor code optimizations

penguinolog:
* `sum`, `tuple` and `min` accept comprehensions
* `val is not None and val != ""` can be simplified

##### Checklist

### DIFF
--- a/urwid/numedit.py
+++ b/urwid/numedit.py
@@ -332,7 +332,7 @@ class FloatEdit(NumEdit):
             preserve_significance = preserveSignificance
 
         val = ""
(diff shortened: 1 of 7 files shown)


urwid/numedit.py
```diff
@@ -332,7 +332,7 @@ def __init__(
             preserve_significance = preserveSignificance
 
         val = ""
-        if default is not None and default != "":
```
wardi, line 335: I think we want to allow `default=0` here

wardi (approved):

PR #679, opened 2023-11-24
Replace deprecated use of MultiError with exceptiongroup

DRMacIver:
##### Checklist

##### Description:

This pull request fixes a deprecation warning when running with recent versions of trio (and a TODO comment from the relevant section of the code) by replacing usage of trio's `MultiError.catch` with `exceptiongroup.catch`. It required a little bit of fiddling to get the right semantics, but as far as I can tell this should now be identical behaviour without the deprecation warning (and it passes all the tests)

This does have the possibly slightly undesirable feature of adding a dependency on exceptiongroup for using trio. I think this is mostly harmless - it's a dependency of trio anyway in Python versions prior to `except*` syntax being added. If you want I can do a slightly more complicated version of this pull request that removes the dependency on Python versions recent enough to not need it, but that seemed likely to be more trouble than it was worth compared to just installing the backport unconditionally.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ classifiers = {file = ["classifiers.txt"]}
 [project.optional-dependencies]
 glib = ["PyGObject"]
 tornado = ["tornado"]
-trio = ["trio", "exceptiongroup"]
+trio = ["trio>=0.22.0", "exceptiongroup"]
 twisted = ["twisted"]
 zmq = ["zmq"]
 # for lcd_display
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ classifiers = {file = ["classifiers.txt"]}
 [project.optional-dependencies]
 glib = ["PyGObject"]
 tornado = ["tornado"]
-trio = ["trio"]
+trio = ["trio", "exceptiongroup"]
 twisted = ["twisted"]
 zmq = ["zmq"]
 # for lcd_display
--- a/urwid/event_loop/trio_loop.py
+++ b/urwid/event_loop/trio_loop.py
@@ -27,6 +27,7 @@ from __future__ import annotations
 
 import typing
 
+import exceptiongroup
 import trio
 
 from .abstract_loop import EventLoop, ExitMainLoop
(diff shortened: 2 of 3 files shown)


urwid/event_loop/trio_loop.py
```diff
@@ -142,8 +143,7 @@ def run(self) -> None:
 
         emulate_idle_callbacks = _TrioIdleCallbackInstrument(self._idle_callbacks)
 
-        # TODO(Aleksei): trio.MultiError is deprecated in favor of exceptiongroup package usage and `Except *`
-        with trio.MultiError.catch(self._handle_main_loop_exception):
+        with exceptiongroup.catch({BaseException: self._handle_main_loop_exception}):
```
penguinolog, line 146: This change needs lower constraints for `trio` version

penguinolog (changes requested):
Small changes needed and LGTM

urwid/event_loop/trio_loop.py
penguinolog: Personally I prefer constraints on requirements, because support of unlimited legacy versions is an unlimited pain.

DRMacIver: Yup, I think this makes more sense too. I just wasn't sure what your preferences were. I've now pushed a version with a constraint on the earliest version of trio where this should work.

pyproject.toml
```diff
@@ -47,7 +47,7 @@ classifiers = {file = ["classifiers.txt"]}
 [project.optional-dependencies]
 glib = ["PyGObject"]
 tornado = ["tornado"]
-trio = ["trio"]
+trio = ["trio>=0.22.0", "exceptiongroup"]
```
penguinolog, line 50: Please also add `exceptiongroup` to the functional tests requirements. Without it tests fail

penguinolog (changes requested):

pyproject.toml
DRMacIver: Whoops, sorry. I'd only run the tests on Python 3.10 so hadn't noticed that they failed on later Python versions. Now pushed a fix.

penguinolog (approved):

PR #691, opened 2023-12-08
Feature: switch from `select.select` to `selectors`

penguinolog:
* Use `DefaultSelector` as automatically selected most efficient.
* Use context manager for not holding resources.
* `InterruptedError` is handled by selector (and since python 3.5 by `select.select`, see PEP 475)

##### Checklist


urwid/event_loop/select_loop.py
```diff
@@ -156,31 +156,31 @@ def _loop(self) -> None:
...
         A single iteration of the event loop
         """
-        fds = list(self._watch_files)
         tm: float | Literal["idle"] | None = None
 
-        if self._alarms or self._did_something:
-            timeout = 0.0
+        with selectors.DefaultSelector() as selector:
+            for fd, callback in self._watch_files.items():
+                selector.register(fd, selectors.EVENT_READ, callback)
```
penguinolog, line 163: use modern api and parse ` self._watch_files` once in iteration

urwid/event_loop/select_loop.py
```diff
@@ -191,6 +191,6 @@ def _loop(self) -> None:
                 alarm_callback()
                 self._did_something = True
 
-        for fd in ready:
-            self._watch_files[fd]()
+        for record in ready:
+            record.data()
```
penguinolog, line 195: we put callback here

wardi (approved):
reduces line count and probably makes SelectEventLoop perform better on linux, so that's great. This doesn't help with pipes not being selectable on windows though, right? Do all apps that want to be cross-platform need to be written to use sockets?

penguinolog:
> reduces line count and probably makes SelectEventLoop perform better on linux, so that's great. This doesn't help with pipes not being selectable on windows though, right? Do all apps that want to be cross-platform need to be written to use sockets?

Unfortunately, yes - only sockets under windows or use some gevent-like approach.
Technically we can use primitives with exchange without `select` for some operations, it was proposed long time ago. It will be not global solution, but reduce pain

Issue #690, opened 2023-12-07
[BUG]On Windows 11 detected encoding is not unicode.


penguinolog:
##### Description:
Real console encoding on windows 11 is UTF-8.
Detected console encoding is cp125*

##### Affected versions (if applicable)
- [x] `master` branch (specify commit)
- [x] Latest stable version from `pypi`

##### Expected/actual outcome
Encoding detected/enforced as UTF-8 and most examples can be executed correctly without modification

PR #692, opened 2023-12-08
Feature: support windows-curses for curses under windows

### LINKED ISSUE #447: partial native Windows support via windows-curses

penguinolog:
* Add `curses` optional dependencies group
* Default windows-curses require UTF-8 encoding to work

### Due to issue #690 not adding windows instructions to the manual

Fix #447

Successfully tested under windows 11 with enforced UTF-8 encoding:
```python
import urwid

urwid.set_encoding("utf-8")
```

##### Checklist

### DIFF
--- a/tests/test_event_loops.py
+++ b/tests/test_event_loops.py
@@ -160,7 +160,10 @@ except ImportError:
     pass
 else:
 
-    @unittest.skipIf(IS_WINDOWS, "Windows is temporary not supported by TornadoEventLoop.")
+    @unittest.skipIf(
+        IS_WINDOWS,
+        "Windows is temporary not supported by TornadoEventLoop due to race conditions.",
+    )
     class TornadoEventLoopTest(unittest.TestCase, EventLoopTestMixin):
         def setUp(self):
             from tornado.ioloop import IOLoop
@@ -176,7 +179,10 @@ except ImportError:
     pass
 else:
 
-    @unittest.skipIf(IS_WINDOWS, "Windows is temporary not supported by TwistedEventLoop.")
+    @unittest.skipIf(
+        IS_WINDOWS,
+        "Windows is temporary not supported by TwistedEventLoop due to race conditions.",
+    )
     class TwistedEventLoopTest(unittest.TestCase, EventLoopTestMixin):
         def setUp(self):
             self.evl = urwid.TwistedEventLoop()
@@ -283,6 +289,10 @@ except ImportError:
     pass
 else:
 
+    @unittest.skipIf(
+        IS_WINDOWS,
+        "Windows is temporary not supported by TrioEventLoop due to race conditions.",
+    )
     class TrioEventLoopTest(unittest.TestCase, EventLoopTestMixin):
         def setUp(self):
             self.evl = urwid.TrioEventLoop()
--- a/urwid/escape.py
+++ b/urwid/escape.py
@@ -24,6 +24,7 @@ Terminal Escape Sequences for input and display
 
 from __future__ import annotations
 
+import os
 import re
 from collections.abc import MutableMapping, Sequence
 
@@ -35,6 +36,8 @@ except ImportError:
 # NOTE: because of circular imports (urwid.util -> urwid.escape -> urwid.util)
 # from urwid.util import is_mouse_event -- will not work here
 import urwid.util
+
+IS_WINDOWS = os.name == "nt"
 
 within_double_byte = str_util.within_double_byte
 
@@ -428,6 +431,10 @@ _keyconv = {
     350: "5",  # on numpad
     360: "end",
 }
+
+if IS_WINDOWS:
+    _keyconv[351] = "shift tab"
+    _keyconv[358] = "end"
 
 
 def process_keyqueue(codes: Sequence[int], more_available: bool) -> tuple[list[str], Sequence[int]]:
--- a/classifiers.txt
+++ b/classifiers.txt
@@ -6,6 +6,7 @@ License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)
 Operating System :: POSIX
 Operating System :: Unix
 Operating System :: MacOS :: MacOS X
+Operating System :: Microsoft :: Windows
 Topic :: Software Development :: Libraries :: Python Modules
 Topic :: Software Development :: Widget Sets
 Programming Language :: Python :: 3
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,6 +97,7 @@ exclude_lines = [
 
   # Exclude methods marked as abstract
   "@abstractmethod",
+  "@abc.abstractmethod",
 
   # Exclude import statements
   "^from\b",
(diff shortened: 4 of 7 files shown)


urwid/escape.py
```diff
@@ -429,6 +432,10 @@ def read_cursor_position(self, keys, more_available: bool):
     360: "end",
 }
 
+if IS_WINDOWS:
+    _keyconv[351] = "shift tab"
+    _keyconv[358] = "end"
```
penguinolog, line 437: works for me

tests/test_event_loops.py
```diff
@@ -160,7 +160,10 @@ def test_error(self):
     pass
 else:
 
-    @unittest.skipIf(IS_WINDOWS, "Windows is temporary not supported by TornadoEventLoop.")
+    @unittest.skipIf(
+        IS_WINDOWS,
+        "Windows is temporary not supported by TornadoEventLoop due to race conditions.",
```
penguinolog, line 165: stable enough, maybe reason to enable some atomic tests before issue will be fully fixed

PR #699, opened 2023-12-11
Documentation: explain Windows support

penguinolog:
* Fix README.rst validation target in tox.ini

##### Checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,8 @@ commands =
 
 [testenv:readme]
 deps =
-  twine build
+  twine
+  build
 commands =
   python -m build
   twine check {toxinidir}/dist/*
--- a/README.rst
+++ b/README.rst
@@ -63,6 +63,8 @@ Windows support notes
 
     pip install urwid[curses]
 
+* Window resize is not fully functional without keyboard/mouse input if RawDisplay used.
+* CursesDisplay decode limited scope of modifier keys combinations.
 * Only UTF-8 mode is supported.
 
 
--- a/README.rst
+++ b/README.rst
@@ -63,8 +63,7 @@ Windows support notes
 
     pip install urwid[curses]
 
-* For Windows OS default encoding depends on python binary version
-  and normally automatically set to the UTF-8.
+* Only UTF-8 mode is supported.
 
 
 Testing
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -12,7 +12,8 @@ commands =
 
 [testenv:readme]
 deps =
-  twine build
```
penguinolog, line 15: inline threated as command instead of list

wardi:
I'm still a little confused about windows support.
- does it depend on WSL?
- does it only work in powershell or does regular cmd work too?
- is raw_display supported or only the curses display module?
- what are the remaining issues with Unicode and supporting other encodings?

Ages ago when I tested on cygwin it seemed to mostly work in that environment/terminal but under `cmd` there was no resizing support. Mingw64 might be another option (cygwin replacement?) for windows support with fewer limitations, I've found it to be a nice way to run bash/ssh when I'm stuck on windows.

penguinolog:
> I'm still a little confused about windows support.
> 
> * does it depend on WSL?
> * does it only work in powershell or does regular cmd work too?
> * is raw_display supported or only the curses display module?
> * what are the remaining issues with Unicode and supporting other encodings?
> 
> Ages ago when I tested on cygwin it seemed to mostly work in that environment/terminal but under `cmd` there was no resizing support. Mingw64 might be another option (cygwin replacement?) for windows support with fewer limitations, I've found it to be a nice way to run bash/ssh when I'm stuck on windows.

1. It not depend on WSL. WSL is linux under windows (literally), it had only issue with RawDisplay due to WSL magic.
2. It works both with `cmd` and `powershell`. Actual `master` branch of `urwid` should execute most UI examples (except with Terminal widget).
3. Supported RawDisplay (we have 2 different versions of RawDisplay: normal and windows with low level Terminal API calls) and curses via windows-curses library (adopted patch from related issue)
4. Windows terminal for a long times unicode and not unicode in the same time - it depends on how data is sent to terminal, as result encoding detection will be always wrong for it.
   If python compiled with unicode support (for windows it's technically always) - it uses Unicode IO and terminal codepage settings does not matter.
   Library windows-curses is distributed compiled for unicode and usage of anything else will cause exceptions (or need to manually compile it from scratch with disabled unicode support).

wardi:
That's great, thanks. So maybe we can also say that under Windows unicode mode is the only one supported.

Issue #702, opened 2023-12-11
[Feature][Windows] Extend modifier keys handling in CursesDisplay under windows.


penguinolog:
At current moment combinations with <ctrl> and <alt> are almost not decoded to human-friendly events and numeric numbers returned instead.
Looks like generator expression approach should allow to fix this issue with minimum amount of code.

penguinolog:
PR #705 fixed main issue

Issue #700, opened 2023-12-11
[BUG][Windows] Terminal window resize not handled by RawDisplay without keypress/mouse event


penguinolog:
##### Description:
Terminal window (`cmd.exe`/`powershell`) resize not causing screen invalidation and re-drawing without keypress/mouse event.
CursesDisplay handles event correctly

##### Affected versions (if applicable)
- [X] `master` branch (specify commit)
- [ ] Latest stable version from `pypi`
- [ ] Other (specify source)

penguinolog:
Related #575

PR #706, opened 2023-12-13
Fix: Phantom NULL keys codes from modifier keys on Win32 RawDisplay

penguinolog:
* For `windows-curses` KEY_MOUSE is 539, btw signals work buggy: in case of fast click, "press" may be handled and "release" missed
* Use max 2 extra wait cycles if multiple resize commands are received. This will allow adequate terminal behavior without CPU overload.
* Explicit close resize command sockets on RawDisplay instance delete
* `get_input_descriptors` method is 90% common for both POSIX and Win32 implementations -> move to the base class and overload
* Win32 `EventType` -> use `IntFlag` type (supported since Python 3.6)

Partial #700
Partial #702

##### Checklist

### DIFF
--- a/urwid/_raw_display_base.py
+++ b/urwid/_raw_display_base.py
@@ -52,6 +52,7 @@ class Screen(BaseScreen, RealTerminal):
         terminal.
         """
         super().__init__()
+        self._partial_codes: list[int] = []
         self._pal_escape = {}
         self._pal_attrspec = {}
         signals.connect_signal(self, UPDATE_PALETTE_ENTRY, self._on_update_palette_entry)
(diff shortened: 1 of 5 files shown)


urwid/_raw_display_base.py
```diff
@@ -262,18 +267,16 @@ def get_input(self, raw_keys: bool = False) -> list[str] | tuple[list[str], list
 
         # Avoid pegging CPU at 100% when slowly resizing
         if keys == ["window resize"] and self.prev_input_resize:
```
penguinolog, line 269: technically this method is not used due to `hook_event_loop` API availability, but looks like kept for compatibility

urwid/_win32.py
```diff
@@ -116,12 +117,12 @@ class INPUT_RECORD(Structure):
     _fields_: typing.ClassVar[list[tuple[str, type]]] = [("EventType", WORD), ("Event", Event)]
 
 
-class EventType:
-    FOCUS_EVENT = 0x0010
+class EventType(enum.IntFlag):
     KEY_EVENT = 0x0001
```
penguinolog, line 121: order incrementally for readability

PR #708, opened 2023-12-15
Feature: Add internal logging for behavioral debug

penguinolog:
* Use standard python logging
* Use nested loggers for possibility to set atomic logging
* Add `__repr__` to the part of critical types
* In screen event loop with `selectors` usage: use IO objects instead of descriptors if possible: get human-friendly output in logs

* `MainLoop` methods`watch_pipe` / `remove_watch_pipe` are not available under Windows

##### Checklist

### DIFF
--- a/urwid/_posix_raw_display.py
+++ b/urwid/_posix_raw_display.py
@@ -27,6 +27,7 @@ from __future__ import annotations
 import contextlib
 import fcntl
 import functools
+import logging
 import os
 import signal
 import struct
@@ -43,6 +44,7 @@ from . import _raw_display_base
 
 if typing.TYPE_CHECKING:
     import io
+    import socket
     from collections.abc import Callable
     from types import FrameType
 
@@ -64,6 +66,7 @@ class Screen(_raw_display_base.Screen):
             and `end paste` keystrokes when the user pastes text.
         """
         super().__init__(input, output)
+        self.logger = logging.getLogger(__name__).getChild(self.__class__.__name__)
         self.gpm_mev: Popen | None = None
         self.gpm_event_pending: bool = False
         self.bracketed_paste_mode = bracketed_paste_mode
@@ -72,6 +75,14 @@ class Screen(_raw_display_base.Screen):
         self._prev_sigcont_handler = None
         self._prev_sigtstp_handler = None
         self._prev_sigwinch_handler = None
+
+    def __repr__(self) -> str:
+        return (
+            f"<{self.__class__.__name__}("
+            f"input={self._term_input_file}, "
+            f"output={self._term_output_file}, "
+            f"bracketed_paste_mode={self.bracketed_paste_mode})>"
+        )
 
     def _sigwinch_handler(self, signum: int, frame: FrameType | None = None) -> None:
         """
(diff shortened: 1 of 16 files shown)


urwid/_posix_raw_display.py
```diff
@@ -371,7 +382,7 @@ def get_cols_rows(self) -> tuple[int, int]:
         y, x = super().get_cols_rows()
         with contextlib.suppress(OSError):  # Term size could not be determined
             if hasattr(self._term_output_file, "fileno"):
-                buf = fcntl.ioctl(self._term_output_file.fileno(), termios.TIOCGWINSZ, " " * 4)
```
penguinolog, line 374: This is buggy code: ioctl expect Buffer protocol, which includes `bytes`, but not `str`

urwid/_raw_display_base.py
```diff
@@ -37,27 +38,27 @@
...
     from urwid import Canvas, EventLoop
 
-    class FileLikeObj(typing.Protocol):
-        def fileno(self) -> int:
-            ...
-
 
 class Screen(BaseScreen, RealTerminal):
-    def __init__(self, input: FileLikeObj, output: FileLikeObj):  # noqa: A002
+    def __init__(self, input: io.IOBase, output: io.IOBase):  # noqa: A002
```
penguinolog, line 51: Use full type

urwid/_win32_raw_display.py
```diff
@@ -171,7 +173,7 @@ def wrapper() -> tuple[list[str], typing.Any] | None:
                 return self.parse_input(event_loop, callback, self.get_available_raw_input())
 
         fds = self.get_input_descriptors()
-        handles = [event_loop.watch_file(fd, wrapper) for fd in fds]
+        handles = [event_loop.watch_file(fd if isinstance(fd, int) else fd.fileno(), wrapper) for fd in fds]
```
penguinolog, line 176: Do not break EventLoop API

urwid/event_loop/main_loop.py
```diff
@@ -242,66 +244,70 @@ def remove_alarm(self, handle) -> bool:
...
+            Close the read end of the pipe and remove the watch created by
+            :meth:`watch_pipe`. You are responsible for closing the write end of
+            the pipe.
+
+            Returns ``True`` if the watch pipe exists, ``False`` otherwise
+            """
+            try:
+                watch_handle, pipe_rd = self._watch_pipes.pop(write_fd)
+            except KeyError:
+                return False
```
wardi, line 305: This is mainly used to trigger a screen update from a signal (like the terminal resize) or from another process/thread. On windows we need something like this, maybe a socket pair connected over loopback?

penguinolog: https://github.com/urwid/urwid/blob/master/urwid/_win32_raw_display.py#L247 ->
https://github.com/urwid/urwid/blob/master/urwid/_win32_raw_display.py#L165C77-L165C94 ->
https://github.com/urwid/urwid/blob/master/urwid/_raw_display_base.py#L99

And `socket` instance is used for sending `b"R"` to trigger update.
Unfortunately it's not enough, so issue #700 created and in process of debugging, when I have physical access to the Windows machine.

Issue #575, opened 2023-06-15
PR #541 is making the toot client stall at startup until a keyboard or mouse event is received


danschwarz:
##### Description:
toot relies on asyncio and does a lot of async network access, especially at startup.  

When running with Urwid release 2.1.2 up to master right before #541 was integrated, starting up toot takes about 6sec to get from title screen to fully loaded timeline, and it does this without any user input.

When running on urwid master,  350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f or later, startup stalls on the title screen showing "Loading..." indefinitely if no keyboard/mouse activity.  If there is keyboard or mouse activity at startup, toot takes about 6sec to get from title screen to fully loaded timeline, as above.

I'm not sure whether this is an issue with the PR, or an issue with how toot is using asyncio;  I'm not very familiar with Python asyncio details.  Including @ihabunek as he may have some insight.

##### Affected versions (if applicable)
- [X] `master` branch (specify commit) 350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f 
- [ ] Latest stable version from `pypi`
- [ ] Other (specify source)

##### Steps to reproduce (if applicable)

You want to test against a [special version of toot](https://github.com/danschwarz/toot/tree/urwid_test2).  This is the current master with one line of code commented out - it removes the BigText name/version number display on the title screen (you'll need my version to run without crashing on Urwid master versions prior to integration of #555 )

##### Expected/actual outcome

When run with Urwid 2.1.2 release, or any version of master prior to 350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f , toot does not stall, it proceeds from title screen  to the fully loaded timeline within 6 seconds.

When run with 350ee5c47ff565d3b0d25b1c3cbddbfb2a0a2a4f  or later versions of master, toot stalls at "loading..." indefinitely, until a keyboard/mouse event is received.

danschwarz:
Updated the title and description to reflect what's really going on here.

penguinolog:
Looks like I'll have to return old behavior and put current asyncio logic for more investigation

danschwarz:
Thanks for looking into this.

penguinolog:
I agree, but at this moment looks like screen not redrawed if no input events (and widgets invalidated). This is blocker

wardi:
[...]

PR #712, opened 2023-12-18
Add `add_run_in_executor` helper to the event loop

penguinolog:
Implemented only for:
* `asyncio`
* `tornado` (`asyncio` as backend since 5.0)
* `select` (call executor directly)
* `glib` (call executor directly)
* `zmq` (call executor directly)

event loops at this moment

* Rework unittests for event loop for maintainability

 Fix: #575

##### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ classifiers = {file = ["classifiers.txt"]}
 [project.optional-dependencies]
 curses = ["windows-curses;sys_platform=='win32'"]
 glib = ["PyGObject"]
-tornado = ["tornado"]
+tornado = ["tornado>=5.0"]
 trio = ["trio>=0.22.0", "exceptiongroup"]
 twisted = ["twisted"]
 zmq = ["zmq"]
--- a/docs/manual/mainloop.rst
+++ b/docs/manual/mainloop.rst
@@ -59,9 +59,9 @@ Event Loops
 
 Urwid's event loop classes handle waiting for things for the
 :class:`MainLoop`. The different event loops allow you to
-integrate with Twisted_, Glib_, Tornado_, Asyncio_ libraries,
-or use a simple ``select``-based
-loop. Event loop classes abstract the particulars of waiting for input and
+integrate with Asyncio_, Twisted_, Glib_, Tornado_, ZMQ_ libraries,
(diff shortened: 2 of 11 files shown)


penguinolog:
@danschwarz @ihabunek see added test as example for API to use for update calls from executors and callbacks usage

pyproject.toml
```diff
@@ -47,7 +47,7 @@ classifiers = {file = ["classifiers.txt"]}
 [project.optional-dependencies]
 curses = ["windows-curses;sys_platform=='win32'"]
 glib = ["PyGObject"]
-tornado = ["tornado"]
```
penguinolog, line 50: <5.0 is not supported since version 2.2.0

penguinolog:
Thecnically it was possible to just expose `_loop` from Asyncio/Tornado event loops, but it will be not so comfortable in IDE's and static checking.
Public API is almost the same (small deviations).

wardi:
Maybe an update for the docs/examples to indicate when and how to use run_in_executor?

wardi:
So how does this change a program that would have called the MainLoop.run method?

penguinolog:
> So how does this change a program that would have called the MainLoop.run method?

Change is needed only if threading is used: by design `asyncio` and `tornado` should call executors using `run_in_executor`. Normally it's associated with event loop blocking, but in our case it associated with not triggering callback set from other thread if nothing happens in main thread (no input and no scheduled event).
As I commented bug, instead of calling `executor.submit` (where executor is `concurrent.futures.ThreadPoolExecutor` or `concurrent.futures.ProcessPoolExecutor`) need to call `evl.run_in_executor` where `evl` is our EventLoop.
Also specially explained `TwistedEventLoop` case (it implements it's own thread pool and will crash in case of usage external executors)

wardi:
Do you think it's worth adding a small example program that demonstrates e.g. using urwid with a worker thread, asyncio and run_in_executor?

penguinolog:
> Do you think it's worth adding a small example program that demonstrates e.g. using urwid with a worker thread, asyncio and run_in_executor?

We can add fixed program from bug, but it's a special case IMHO. We can go too deep in examples.

wardi (approved):

PR #715, opened 2023-12-19
Force `SelectorEventLoop` for asyncio under Windows if event loop is not set

penguinolog:
* Document `SelectorEventLoop` is only supported under windows.
* Document `uvloop` is officially not supported.

##### Checklist

### DIFF
--- a/docs/manual/mainloop.rst
+++ b/docs/manual/mainloop.rst
@@ -111,7 +111,15 @@ This event loop integrates with the asyncio module in Python.
     Use instead method :method:`run_in_executor` of :class:`AsyncioEventLoop`,
     which forward API of the same method from asyncio Event Loop run_in_executor_.
 
-.. _run_in_executor: https://docs.python.org/3.11/library/asyncio-eventloop.html#asyncio.loop.run_in_executor
+.. warning::
+
+    For input handling, selectors based logic is used.
+    Under Windows OS only `SelectorEventLoop`_ is supported.
+
+.. note:: `uvloop` event loop is not supported officially.
+
+.. _run_in_executor: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor
+.. _SelectorEventLoop: https://docs.python.org/3/library/asyncio-eventloop.html?highlight=selectoreventloop#asyncio.SelectorEventLoop
 
 .. seealso::
 
--- a/urwid/event_loop/asyncio_loop.py
+++ b/urwid/event_loop/asyncio_loop.py
@@ -69,16 +69,16 @@ class AsyncioEventLoop(EventLoop):
         self.logger = logging.getLogger(__name__).getChild(self.__class__.__name__)
         if loop:
             self._loop: asyncio.AbstractEventLoop = loop
-            self.__event_loop_policy_altered: bool = False
-            self.__original_event_loop_policy: asyncio.AbstractEventLoopPolicy | None = None
+            self._event_loop_policy_altered: bool = False
+            self._original_event_loop_policy: asyncio.AbstractEventLoopPolicy | None = None
         else:
-            self.__original_event_loop_policy = asyncio.get_event_loop_policy()
-            if IS_WINDOWS and not isinstance(self.__original_event_loop_policy, asyncio.WindowsSelectorEventLoopPolicy):
+            self._original_event_loop_policy = asyncio.get_event_loop_policy()
+            if IS_WINDOWS and not isinstance(self._original_event_loop_policy, asyncio.WindowsSelectorEventLoopPolicy):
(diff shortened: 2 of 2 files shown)


docs/manual/mainloop.rst
```diff
@@ -111,7 +111,15 @@ This event loop integrates with the asyncio module in Python.
     Use instead method :method:`run_in_executor` of :class:`AsyncioEventLoop`,
     which forward API of the same method from asyncio Event Loop run_in_executor_.
 
-.. _run_in_executor: https://docs.python.org/3.11/library/asyncio-eventloop.html#asyncio.loop.run_in_executor
+.. warning::
```
penguinolog, line 114: make warning, because we do not want to read bugs "I used incompatible event loop and it crashed with AttributeError"

docs/manual/mainloop.rst
```diff
@@ -111,7 +111,15 @@ This event loop integrates with the asyncio module in Python.
...
+
+    For input handling, selectors based logic is used.
+    Under Windows OS only `SelectorEventLoop`_ is supported.
+
+.. note:: `uvloop` event loop is not supported officially.
```
penguinolog, line 119: I tried to test it and it successfully failed all unittests

urwid/event_loop/asyncio_loop.py
```diff
@@ -65,6 +67,9 @@ def __init__(self, *, loop: asyncio.AbstractEventLoop | None = None, **kwargs) -
         if loop:
             self._loop: asyncio.AbstractEventLoop = loop
         else:
+            if IS_WINDOWS:
+                asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
wardi, line 71: setting a global policy as a side-effect of creating an instance of AsyncioEventLoop is surprising. At a minimum this should be warned about in the class docstring

penguinolog: Added log record and restore of default in destructor if changed

urwid/event_loop/asyncio_loop.py
```diff
@@ -40,6 +41,7 @@
     _T = typing.TypeVar("_T")
 
 __all__ = ("AsyncioEventLoop",)
+IS_WINDOWS = sys.platform == "win32"
 
 
 class AsyncioEventLoop(EventLoop):
```
wardi, line 47: Please warn about constructor/destructor side-effects on windows in the docstring so that users aren't surprised

urwid/event_loop/asyncio_loop.py
```diff
@@ -57,14 +59,20 @@ class AsyncioEventLoop(EventLoop):
...
     """
 
-    _we_started_event_loop = False
-
     def __init__(self, *, loop: asyncio.AbstractEventLoop | None = None, **kwargs) -> None:
         super().__init__()
         self.logger = logging.getLogger(__name__).getChild(self.__class__.__name__)
         if loop:
             self._loop: asyncio.AbstractEventLoop = loop
+            self.__event_loop_policy_altered: bool = False
```
wardi, line 67: double-underscore will munge the name and make it more difficult to update/access from a subclass. Can we use a single underscore instead?

urwid/event_loop/asyncio_loop.py
```diff
@@ -73,6 +81,10 @@ def __init__(self, *, loop: asyncio.AbstractEventLoop | None = None, **kwargs) -
         self._idle_handle: int = 0
         self._idle_callbacks: dict[int, Callable[[], typing.Any]] = {}
 
+    def __del__(self) -> None:
+        if self.__event_loop_policy_altered:
+            asyncio.set_event_loop_policy(None)  # Restore default event loop policy
```
wardi, line 86: should we capture the previous policy and restore it instead of setting back to the default?

urwid/event_loop/asyncio_loop.py
penguinolog: Better to not allow such changes from outside

wardi: I disagree because there are no real private/protected variables in python so it's better to indicate your intent with a single underscore than to try and fake it with a munged name.

If someone wants to override our values from their code this won't prevent them and there's no reason for us to assume we know better what they need in their particular environment/application.

urwid/event_loop/asyncio_loop.py
penguinolog: done

urwid/event_loop/asyncio_loop.py
penguinolog: done

urwid/event_loop/asyncio_loop.py
penguinolog: not fully agree, but removed mangle

wardi (approved):

PR #716, opened 2023-12-19
Add type annotations for TrioEventLoop

penguinolog:
##### Checklist

### DIFF
--- a/urwid/event_loop/trio_loop.py
+++ b/urwid/event_loop/trio_loop.py
@@ -34,7 +34,12 @@ import trio
 from .abstract_loop import EventLoop, ExitMainLoop
 
 if typing.TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


wardi (approved):
just curious about putting imports in `if typing.TYPE_CHECKING` blocks: why doesn't this cause a NameError on the definitions when running normally?

penguinolog:
> just curious about putting imports in `if typing.TYPE_CHECKING` blocks: why doesn't this cause a NameError on the definitions when running normally?

the first import is `from __future__ import annotations`.
It changes type annotations behavior: annotations are handled as text and interpreted only during type checking or on manual explicit call via `inspect` API (it's a totally special case).
Such behavior fixes forward annotations and cyclic imports, where import was used only for typing purposes.

Issue #703, opened 2023-12-11
[Feature][API] get rid of `hasattr` checking for parent widget API calls


penguinolog:
Methods `get_pref_col`, `move_cursor_to_coords` and `mouse_event` is checked for existency and causing behavior change.
Need to normalize API to avoid dependency on `hasattr` and `getattr` calls.

PR #718, opened 2023-12-20
Implement default keypress/mouse_event in Widget

penguinolog:
* log if widget not selectable
* `keypress` -> return key (unhandled)
* `mouse_event` -> return `False` (unhandled)

Partial #703

##### Checklist

### DIFF
--- a/urwid/widget/widget.py
+++ b/urwid/widget/widget.py
@@ -21,6 +21,7 @@
 from __future__ import annotations
 
 import functools
+import logging
 import typing
 import warnings
 from operator import attrgetter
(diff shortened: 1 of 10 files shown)


urwid/widget/widget.py
```diff
@@ -740,12 +799,16 @@ def move_cursor_to_coords(self) -> Callable[[[tuple[()] | tuple[int] | tuple[int
         def rows(self, size: tuple[int], focus: bool = False) -> int:
             return get_delegate(self).rows(size, focus=focus)
 
-        @property
         def mouse_event(
```
penguinolog, line 802: now implemented in base class -> exists always

wardi:
historically there was no requirement to inherit from Widget, so this would be worth mentioning in the changelog/release announcement

penguinolog:
Looks like I'll postpone this PR as potentially API breaking

penguinolog:
Replaced with non-breaking implementation

PR #720, opened 2023-12-27
Feature: Use module path + class name for loggers and init once

penguinolog:
* Add `logger` to all `Widget` instances (if properly subclassed)

##### Checklist

### DIFF
--- a/urwid/display_common.py
+++ b/urwid/display_common.py
@@ -37,7 +37,6 @@ if typing.TYPE_CHECKING:
 
     from urwid import Canvas
 
-LOGGER = logging.getLogger(__name__)
 IS_WINDOWS = sys.platform == "win32"
 
 if not IS_WINDOWS:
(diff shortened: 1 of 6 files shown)


urwid/display_common.py
```diff
@@ -1004,7 +1003,7 @@ class BaseScreen(metaclass=BaseMeta):
     def __init__(self) -> None:
         super().__init__()
 
-        self.logger = LOGGER.getChild(self.__class__.__name__)
```
wardi, line 1007: this logger could be part of the class instead of the instance so it doesn't have to be tracked in every single widget

penguinolog: It will require metaclass usage, which is a bit overkill

wardi: It could also be a Widget property or method that returns the logger

penguinolog: Technically once logger initializer, it stored in `logging.Manager`, so all instances of the same `Widget` class uses the same `Logger`.
Adding property may break downstream, because it will cause errors if downstream initialize `logger` itself.

PR #721, opened 2023-12-27
Feature: Implement `keypress` and `mouse_event` default handlers

penguinolog:
* `Widget` class to have default standard handlers
* Produce debug log for non-selectable widgets received events
* Produce `WidgetWarning` if `__init__` of base class not called
* Produce deprecation warnings if no subclassed Widget

##### Checklist

### DIFF
--- a/urwid/widget/widget.py
+++ b/urwid/widget/widget.py
@@ -585,7 +585,7 @@ class Widget(metaclass=WidgetMeta):
                 self.logger.debug(f"keypress sent to non selectable widget {self!r}")
             else:
                 warnings.warn(
-                    f"Widget {self.__class__.__name__} not called 'super().__init__()",
+                    f"Widget {self.__class__.__name__} did not call 'super().__init__()",
                     WidgetWarning,
                     stacklevel=3,
                 )
--- a/urwid/widget/widget.py
+++ b/urwid/widget/widget.py
@@ -582,7 +582,7 @@ class Widget(metaclass=WidgetMeta):
         """
         if not self.selectable():
             if hasattr(self, "logger"):
-                self.logger.debug(f"Widget {self!r} is not selectable")
+                self.logger.debug(f"keypress sent to non selectable widget {self!r}")
             else:
                 warnings.warn(
                     f"Widget {self.__class__.__name__} not called 'super().__init__()",
--- a/urwid/widget/widget.py
+++ b/urwid/widget/widget.py
@@ -761,7 +761,7 @@ class FixedWidget(Widget):
         raise NotImplementedError()
 
 
-def delegate_to_widget_mixin(attribute_name: str):
+def delegate_to_widget_mixin(attribute_name: str) -> type[Widget]:
     """
     Return a mixin class that delegates all standard widget methods
     to an attribute given by attribute_name.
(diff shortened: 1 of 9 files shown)


penguinolog:
This PR should be used instead of #718 : do not break existing code

wardi:
should also update docs if subclassing Widget will be required

penguinolog:
> should also update docs if subclassing Widget will be required

https://github.com/urwid/urwid/blob/master/docs/manual/widgets.rst?plain=1#L656

It is already in documentation for the last 11 years

penguinolog:
Validated `pudb`: fully compatible. With all enabled warnings:
> <string>:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='tst.py' mode='r' encoding='UTF-8'>

(Source is not urwid side)

penguinolog:
Validated `nomadnet`, `s-tui` and `mitmproxy`: no failures

urwid/widget/widget.py
```diff
@@ -576,6 +570,69 @@ def _repr_words(self):
...
+        :param size: See :meth:`Widget.render` for details
+        :type size: tuple[()] | tuple[int] | tuple[int, int]
+        :param key: a single keystroke value; see :ref:`keyboard-input`
+        :type key: str
+        :return: ``None`` if *key* was handled by *key* (the same value passed) if *key* was not handled
+        :rtype: str | None
+        """
+        if not self.selectable():
+            if hasattr(self, "logger"):
+                self.logger.debug(f"Widget {self!r} is not selectable")
```
wardi, line 585: Is this a clearer message?
```suggestion
                self.logger.debug(f"keypress sent to non selectable widget {self!r}")
```

urwid/widget/widget.py
```diff
@@ -576,6 +570,69 @@ def _repr_words(self):
...
+            else:
+                warnings.warn(
+                    f"Widget {self.__class__.__name__} not called 'super().__init__()",
```
wardi, line 588:
```suggestion
                    f"Widget {self.__class__.__name__} did not call 'super().__init__()",
```

urwid/widget/widget_decoration.py
```diff
@@ -30,6 +30,17 @@ class WidgetDecoration(Widget):  # "decorator" was already taken
     """
 
     def __init__(self, original_widget: Widget) -> None:
+        # TODO(Aleksei): reduce amount of multiple inheritance usage
+        # Special case: subclasses with multiple inheritance causes `super` call wrong way
```
wardi, line 34: What's an example of this? `super()` MRO should resolve `__init__` call order for diamond inheritance cases

penguinolog: For `LineBox` it's resolved incorrectly.
Technically `LineBox` using circular inheritance from `Widget`, but in this case `super` call in `WidgetDecoration` really calls like from `LineBox` to the `WidgetWrap`:

```
  File "/home/alekseis/PycharmProjects/urwid/urwid/widget/line_box.py", line 124, in __init__
    WidgetDecoration.__init__(self, original_widget)
  File "/home/alekseis/PycharmProjects/urwid/urwid/widget/widget_decoration.py", line 37, in __init__
    super().__init__()
TypeError: WidgetWrap.__init__() missing 1 required positional argument: 'w'
```

wardi (approved):

Issue #225, opened 2017-04-15
LineBox rline not positioned correctly around Text containing symbols from Unicode block Miscellaneous Symbols and Pictographs'


imrek:

amanagr:
@imrek I am facing the same issue, were you able to find any solution?

amanagr:
![screenshot from 2018-04-05 09-21-50](https://user-images.githubusercontent.com/25124304/38346497-153300be-38b3-11e8-8c84-cba2ce8968a5.png)

As shown in the screenshot, the linebox is not aligned properly. Here is the [code](https://github.com/zulip/zulip-terminal/pull/67/files#diff-2dc035187fab6f2d09ad35a111b1e995R137).

amanagr:
But after removing the LineBox the misalignment is still there.
![screenshot from 2018-04-05 10-00-28](https://user-images.githubusercontent.com/25124304/38347314-57002b2a-38b8-11e8-95ae-2e5cfa904f9c.png)

amanagr:
I researched a bit and the problems seem to occur only with [these](https://unicode.org/emoji/charts/full-emoji-list.html) characters. @wardi FYI

eevee:
Odd.  If urwid were unaware that these characters are double-width, I'd expect the box to be drawn too _wide_, not too narrow.

wardi:
urwid assumes they will be rendered double width, when the terminal doesn't it doesn't detect that.

It's kind of a hard problem.

To know how characters are rendered you'd need to send each one to the terminal and get the terminal to report the cursor position back. Then you need to remember all those rendered sizes for each terminal you're rendering to (multiple different terminals could be connected to a single urwid application over ssh for example). Then you need to associate the terminal being rendered to when calculating the layout for each screen (rendered character width affects text wrapping). Then you need to figure out how to do all that without creating an unacceptable lag if the terminal is on a remote connection and slow to respond.

Or maybe there's a quick hack that would do better than the dumb static table of character widths we're using right now without requiring all the changes above. Suggestions welcome!

wardi:
quick hack idea: we could keep our static table but force a cursor reposition after any double-width characters are rendered.

eliasdorneles:
Would it be possible to build a better static table?
Maybe using [unicodedata.east_asian_width()](https://docs.python.org/2/library/unicodedata.html#unicodedata.east_asian_width)?

It seems to bring the right info for the character from this example:

```
>>> import unicodedata
[...]

PR #744, opened 2024-01-12
Update unicode table to the version 15.1.0

penguinolog:
Partially fix render of emoji

##### Checklist

Partial #225


source/str_util.c
```diff
@@ -30,47 +30,828 @@
 #define ENC_WIDE 2
 #define ENC_NARROW 3
 
-static int widths_len = 2*38;
+static int widths_len = 2*820;
```
wardi, line 33: Doing a linear search over 820 items might slow. Should we think of using a more efficient data structure or am I overthinking this?

penguinolog: I'm even thinking about depreciation of C code and migration on `wcwidth.
It became global standard, handle multiple unicode versions and since author allow to copy-paste unicode version detect logic - we can better handle different terminals.
Re-implementation of unicode tables in C or migration on different language (like rust) at this moment is too painful

wardi: detecting things about the terminal is tricky because the text calculations are separated from the screen instance. If someone is connecting to an urwid app over ssh like the twisted ssh example how do we know which unicode version detection to use? (there can be many users attached each with their own screen)

wardi: but, using an existing library for the width calculation sounds like a great idea

penguinolog: And simple copy-paste of logic and content (even if separate generated) looks strange.
We can use binary search, it is faster.

penguinolog: I checked: for current release we can use binary search and later we can migrate C part to rust and use upstream libraries for char width both in python and binary.

danschwarz: Re: use of wcwidth - good idea to do so, but I suggest you have a look at a recent issue posted by the author on his github, where he discusses the limits of UNICODE_VERSION https://github.com/jquast/wcwidth/issues/104

the tl dr is that it's challenging to fix unicode version compatibility issues at the app layer, and the way forward is to get terminal emulator authors to stay up to date with the latest unicode version.

penguinolog: > Re: use of wcwidth - good idea to do so, but I suggest you have a look at a recent issue posted by the author on his github, where he discusses the limits of UNICODE_VERSION https://github.com/jquast/wcwidth/issues/104
>

I fully understand that issue, but it also has limitations: in case of widget decorations, incorrect symbols length of decorated widget will produce incorrect decorations length.
Technically we can play "stupid" and go forward with the last unicode version tables, since only few terminals at this moment fully stuck at unicode 12.

> the tl dr is that it's challenging to fix unicode version compatibility issues at the app layer, and the way forward is to get terminal emulator authors to stay up to date with the latest unicode version.

danschwarz: Understood. Makes sense to update the tables. I am curious, why update to v.12 and not the latest v.15.1?

penguinolog: Originally went to the most supported.
After rethinking on weekends went to the 15.1
Technically we have only 1 fully working solution: ask terminal for cursor position after each double width character and add space if it rendered single width, but this solution is extremely expensive and works only with RawDisplay.

danschwarz (approved):
LGTM

ulidtko (changes requested):
A few questions.

1. Why are we duplicating the logic of Python's built-in stdlib? https://docs.python.org/3/library/unicodedata.html#unicodedata.east_asian_width
2. Can we get rid of the C impl?.. It has caused build issues which'd not exist in a pure-python package.
3. Can we compress the table by assuming width 1 for absent entries?

penguinolog:
> A few questions.
> 
> 1. Why are we duplicating the logic of Python's built-in stdlib? https://docs.python.org/3/library/unicodedata.html#unicodedata.east_asian_width
> 2. Can we get rid of the C impl?.. It has caused build issues which'd not exist in a pure-python package.
> 3. Can we compress the table by assuming width 1 for absent entries?

1. Stdlib unicodedata maybe (in most cases) outdated on the end user machine:
```python
>>> import sys; print('Python %s on %s' % (sys.version, sys.platform))
Python 3.11.4 (main, Dec  7 2023, 15:43:41) [GCC 12.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.15.0 -- An enhanced Interactive Python. Type '?' for help.
PyDev console: using IPython 8.15.0
Python 3.11.4 (main, Dec  7 2023, 15:43:41) [GCC 12.3.0] on linux
>>> import unicodedata
>>> unicodedata.unidata_version
'14.0.0'
```

2. Yes, we can. But it will not fix the main issue (unicode data in python, urwid and terminal are different). And now we do not have build issues as using standard toolchain for release.
3. As discussed, we'll migrate to `wcwidth`, which became standard de-facto. It's already using 2 tables instead of 1. It requires also to adopt to "-1" width (urwid do not expect such length).

jquast:
https://docs.python.org/3/library/unicodedata.html#unicodedata.east_asian_width and other functions in unicodedata library are not sufficient for measuring zero-width, combining, VS-16 emoji, or ZWJ emoji sequences. There is a lot of work done in wcwidth library ([update-tables.py](https://github.com/jquast/wcwidth/blob/master/bin/update-tables.py)) to transfer unicode.org data files into efficient lookup tables and make corrections in their categorization that the unicodedata library does not perform.

(I too would like to use the standard python library to measure characters for a terminal, I do hope to work on that this year https://github.com/jquast/wcwidth/issues/94)

A basic example,

>>> l='💪🏿'
>>> [unicodedata.name(z) for z in l]
['FLEXED BICEPS', 'EMOJI MODIFIER FITZPATRICK TYPE-6']
>>> [unicodedata.combining(z) for z in l]
[0, 0]
>>> [unicodedata.east_asian_width(z) for z in l]
['W', 'W']
```
In this case, there is insufficient data in the std unicodedata library measure the width of this emoji,

while,
```
>>> import wcwidth
>>> wcwidth.wcswidth(l)
2
```

**Comment by penguinolog** [comment]
Switched python implementation to the `wcwidth`.
C implementation is untouched.

File: urwid/old_str_util.py [hunk 2]
```diff
@@ -23,71 +23,24 @@
...
-
-# ACCESSOR FUNCTIONS
-
 
 def get_width(o: int) -> Literal[0, 1, 2]:
     """Return the screen column width for unicode ordinal o."""
-    if o in {0xE, 0xF}:
+    width = wcwidth.wcwidth(chr(o), UNICODE_VERSION)
+    if width < 0:
         return 0
```
penguinolog, line 42: urwid using only positive width

penguinolog:
Looks like Unicode problem is raising. I have to merge this PR and later work on proper Unicode handling in the internals.

Issue #393, opened 2020-01-13
urwid-2.1.0 byte-compiling with py2.7 failed due to _async_kw_event_loop.py (py3 only)


clemat:
Following quick chat with @ulidtko on your gitter community, I am opening this Issue for the record. Even thought there's a valid workaround it would be nice to have it fixed upstream.

##### Description:
File "/usr/local/lib/python2.7/site-packages/urwid/_async_kw_event_loop.py", line 164
    async def _alarm_task(self, scope, seconds, callback):
            ^
SyntaxError: invalid syntax

py2 doesn't support async stuff. This is impacting distribution packages (error when building bytecode). 

removing _async_kw_event_loop.py post-extract for py2 is a valid workaround.
Couldn't this be "excluded" via setuptools? Or other ways. So it doesn't impact all distribution packages and ports.

##### Affected versions 
- 2.1.0

penguinolog:
Python 2.7 support ended. New release drop it

PR #748, opened 2024-01-16
LineBox: make side only if side elements present, fix pack

### LINKED ISSUE #337: Gap between LineBoxes in adjacent columns
##### Description:
I have a `Columns` container with two `LineBox`es. They are set so that they should share the dividing vertical line, but there is a gap.

image

Latest stable version from `pypi` (2.0.1)

```
urwid.Columns([
      urwid.LineBox(
        urwid.Text('lol'),
        rline='', trcorner='', brcorner=''
      ),
      urwid.LineBox(
        urwid.Text('wtf'),
      ),
    ])
```

Setting `dividechars` to -1 resolves it but I don't think that should be necessary.

penguinolog:
* BUG: `LineBox().pack` was ignored nonexistent top/bottom. Switch to the backend (Pile) `pack` usage to get proper size
* Semi-bug: in case of no side, empty elements was partially added (top and bottom) with fixed size == 1  (Bug #393 )
* BUG: `set_title` without title was not properly checked for top widget availability and caused exception
* For consistency add right corner with width=0 if no right corner
* Propagate default empty size for `pack` if widgets are optionally FIXED No size will be handled internally

Fix #337

##### Checklist

### DIFF
--- a/tests/test_line_box.py
+++ b/tests/test_line_box.py
@@ -1,0 +1,167 @@
+from __future__ import annotations
+
+import unittest
+
+import urwid
+from urwid.util import get_encoding
+
+
+class LineBoxTest(unittest.TestCase):
+    def setUp(self) -> None:
+        self.old_encoding = get_encoding()
+        urwid.set_encoding("utf-8")
+
+    def tearDown(self) -> None:
+        urwid.set_encoding(self.old_encoding)
+
+    def test_linebox_pack(self):
+        # Bug #346 'pack' Padding does not run with LineBox
+        t = urwid.Text("AAA\nCCC\nDDD")
+        size = t.pack()
+        l = urwid.LineBox(t)
+
+        self.assertEqual(l.pack()[0], size[0] + 2)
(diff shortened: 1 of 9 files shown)


tests/test_line_box.py
```diff
@@ -0,0 +1,187 @@
...
+                ],
+                [line.decode("utf-8") for line in canvas.text],
+            )
+            self.assertIsNone(l.keypress((), "enter"))
+            self.assertTrue(pressed)
+
+        pressed = False
+
+        with self.subTest("No left side elements -> no side"):
+            l = urwid.LineBox(wrapped, tlcorner="", lline="", blcorner="")
```
wardi, line 133: do we need to remove corners when lline is empty?

wardi: I mean it would be a nicer interface if we didn't, like with top and bottom

penguinolog: We can add it to make consistent with top and bottom

penguinolog: Done

wardi (approved):

PR #761, opened 2024-01-22
Fix: columns rendered with a non-first Selectable widget

penguinolog:
* In case of first elements fit, it should be included in the rendered widget.
* Render columns with FIXED mode should be the same as FLOW mode with cols collected from `widget.pack()`

##### Checklist

### DIFF
--- a/urwid/widget/columns.py
+++ b/urwid/widget/columns.py
@@ -4,6 +4,7 @@ import typing
 import warnings
 from itertools import chain, repeat
 
+import urwid
 from urwid.canvas import Canvas, CanvasJoin, CompositeCanvas, SolidCanvas
 from urwid.monitored_list import MonitoredFocusList, MonitoredList
 from urwid.util import is_mouse_press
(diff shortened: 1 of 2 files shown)


penguinolog:
@wardi this pr fixes old BUG, this is not a regression. Validated, that this bug present in 2.1.2

urwid/widget/columns.py
```diff
@@ -691,12 +692,19 @@ def column_widths(self, size: tuple[int] | tuple[int, int], focus: bool = False)
...
+                else:
+                    warnings.warn(f"{w!r} is not a Widget", ColumnsWarning, stacklevel=3)
+                    w_sizing = frozenset((urwid.BOX, urwid.FLOW))
+
+                if Sizing.FIXED in w_sizing:
+                    w_size = ()
+
+                elif Sizing.FLOW in w_sizing:
                     # FIXME: should be able to pack with a different maxcol value
                     w_size = (maxcol,)
```
penguinolog, line 706: here still present part of bug: when we do not know widget size, we assume whole `cols` and element can be skipped.
Unfortunately, I do not see any good to make universal fix.

wardi (approved):

PR #762, opened 2024-01-22
Feature: Partial FIXED support for `Edit` widget

penguinolog:
Full fixed mode will look ugly due to widget size change on input,
but a size collection and render for debug purposes will be very useful.

##### Checklist

### DIFF
--- a/urwid/widget/edit.py
+++ b/urwid/widget/edit.py
@@ -118,6 +118,22 @@ class Edit(Text):
         self.set_mask(mask)
         self._shift_view_to_cursor = False
 
+    def pack(self, size: tuple[int] | tuple[()] | None = None, focus: bool = False) -> tuple[int, int]:
+        """
+        Return the number of screen columns and rows required for this Edit widget to be displayed
+        without wrapping or clipping, as a single element tuple.
+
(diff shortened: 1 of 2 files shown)


urwid/widget/edit.py
```diff
@@ -118,6 +118,20 @@ def __init__(
...
+        without wrapping or clipping, as a single element tuple.
+
+        Due to "SELECTABLE" specific, an extra 1 symbol is reserved in "FIXED" pack mode
+        to prevent clipping for cursor rendering when the widget is focused.
+        """
+        cols, rows = super().pack(size, focus)
+        if size:
+            return cols, rows
+
+        return cols + 1, rows
```
wardi, line 133: Maybe it's fine to not support FIXED mode for Edit. The whole point of the widget is to accept input so it shouldn't be displayed at some exact minimum size.

(also if the edit text or label contains a newline the cols + 1 might be meaningless)

penguinolog: `_sizing = frozenset([Sizing.FLOW])` is not changed, because this PR is only for debugging purposes, when something already went wrong

PR #782, opened 2024-01-31
Unicode: use "target encoding" while transcoding for output

penguinolog:
* apply `Canvas.content` method API normalization and annotation
* * Overloaded methods should accept arguments as base class method
* * Overloaded methods can use own defaults

##### Checklist

### DIFF
--- a/urwid/display/_raw_display_base.py
+++ b/urwid/display/_raw_display_base.py
@@ -590,7 +590,7 @@ class Screen(BaseScreen, RealTerminal):
 
         last_attributes = None  # Default = empty
 
-        output = [escape.HIDE_CURSOR, attr_to_escape(last_attributes)]
+        output: list[str] = [escape.HIDE_CURSOR, attr_to_escape(last_attributes)]
 
         def partial_display() -> bool:
             # returns True if the screen is in partial display mode ie. only some rows belong to the display
@@ -672,7 +672,7 @@ class Screen(BaseScreen, RealTerminal):
                         output.append(escape.SO)
                     last_charset_flag = cs
 
-                output.append(run)
+                output.append(run.decode(encoding, "replace"))
                 first = False
 
             if ins:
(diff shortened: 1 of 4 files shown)


urwid/display/_raw_display_base.py
```diff
@@ -682,25 +682,25 @@ def using_standout_or_underline(a):
...
             self._cy = y
 
         if self._resized:
             # handle resize before trying to draw screen
             return
         try:
-            for line in o:
+            for line in output:
                 if isinstance(line, bytes):
-                    line = line.decode("utf-8", "replace")  # noqa: PLW2901
```
penguinolog, line 703: this buggy line screams: we should review `Canvas` logic and ultimately switch to the Unicode strings in canvas contents.
Currently we have standard chain for 99,9% of widgets str -> bytes -> str (terminal encoding independent).
During transcoding we're using "target encoding", but in fact it useful only in input processing (not everywhere: in windows and with curses we already can have decoded data).

urwid/display/_raw_display_base.py
```diff
@@ -654,22 +654,22 @@ def using_standout_or_underline(a):
...
+                        output.append(escape.SI)
                     elif cs == "U":
-                        o.append(escape.IBMPC_ON)
+                        output.append(escape.IBMPC_ON)
                     else:
-                        o.append(escape.SO)
+                        output.append(escape.SO)
                     lastcs = cs
-                o.append(run)
+                output.append(run.decode(util.get_encoding(), "replace"))
```
wardi, line 672: Hmm, this code was originally working with bytes so that applications could be running with a non-utf-8 encoding without converting to Unicode and back (potentially losing data in the process) but I guess everything in raw display uses Unicode now?

penguinolog: > Hmm, this code was originally working with bytes so that applications could be running with a non-utf-8 encoding without converting to Unicode and back (potentially losing data in the process) but I guess everything in raw display uses Unicode now?

Output in unicode was as minimum in version 2.0 (and even ignored target encoding): https://github.com/urwid/urwid/blob/release-2.0.0/urwid/raw_display.py#L848

penguinolog: @wardi in python 3 we have STDOUT, which is always "Unicode". And it's not always buffered (so we cannot use buffer for output).

Issue #774, opened 2024-01-29
Weird rendering bug on Windows only


mfncooper:
Using a full-screen LineBox as the main window area, I am seeing a weird character in the left-most column of every row. I don't know what the character is - it looks like an outline of a square with tiny lines coming out of top, bottom, left and right. And because of this character in the left-most column, everything else gets shifted off to the right by one character, so I lose the right-most column. There is also one instance of the same character at the bottom right of the screen.

This only happens on Windows. (I am using Windows 10 with Python 3.10.11.) The same code works perfectly on both Mac and Linux. In all cases, I'm using Urwid 2.4.6 installed from PyPI using pip.

I've stripped down the code to a short app that reproduces the problem, and included that here. I've also included a screenshot of the app when running under Windows.

```python
import urwid

class ContentPanel(urwid.WidgetWrap):
    def __init__(self):
        self._text = urwid.Text("Hello, World!")
        self._fill = urwid.Filler(self._text)
        super().__init__(self._fill)

class MainScreen(urwid.WidgetWrap):
    def __init__(self, panel):
        pile = urwid.Pile([('weight', 1, panel)])
        super().__init__(urwid.LineBox(
            pile, title="Main Screen", title_align='center'))

def create_widgets():
    content_panel = ContentPanel()
    main_screen = MainScreen(content_panel)
    frame = urwid.Frame(main_screen)
    return frame

def unhandled_input(key):
    if key == 'q' or key == 'Q':
        raise urwid.ExitMainLoop()
    return False

def run():
    loop = urwid.MainLoop(
        create_widgets(),
        unhandled_input=unhandled_input)
    loop.run()

if __name__ == "__main__":
    run()
```

![RenderBug](https://github.com/urwid/urwid/assets/450625/c2631cdd-f62c-45f7-9f16-a3e71ec22093)

penguinolog:
Maybe Unicode processing issue: consistent Unicode symbol on screenshot.
Need to validate, that "stupid" `print(create_widgets().render((80, 24)))` is correct.
Expect that `print` will work correctly and issue is somewhere in final screen rendering (windows is "special").

mfncooper:
@penguinolog You are correct. I replaced `run()` in the above code with `print(create_widgets().render((80, 24)))` as you suggested, and the output was correct. I also tried writing the `print()` output to a file (as UTF-8, of course) to verify that there are no "invisible" characters, and there are none.

[...]

PR #785, opened 2024-02-01
Windows and WSL: SI/SO/IBMPC_ON/IBMPC_OFF skip

penguinolog:
In the case of Windows terminal, output is always Unicode and escape sequences for charset manipulation are "garbage"

##### Checklist

Partial #774

### DIFF
--- a/urwid/display/_raw_display_base.py
+++ b/urwid/display/_raw_display_base.py
@@ -28,6 +28,7 @@ import abc
 import contextlib
 import functools
 import os
+import platform
 import selectors
 import signal
 import socket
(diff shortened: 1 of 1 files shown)


urwid/display/_raw_display_base.py
```diff
@@ -640,14 +642,17 @@ def using_standout_or_underline(a: AttrSpec | str) -> bool:
...
+
                 if cs != "U":
                     run = run.translate(UNPRINTABLE_TRANS_TABLE)  # noqa: PLW2901
+
                 if first or lasta != a:
                     o.append(attr_to_escape(a))
                     lasta = a
-                if first or lastcs != cs:
+
+                if not (IS_WINDOWS or IS_WSL) and (first or lastcs != cs):
```
penguinolog, line 655: technically it's not needed at all if console is Unicode but need deeper testing.

Issue #801, opened 2024-02-14
Dropping str_util.c


mfncooper:
Today, Urwid creates a crazy number of wheel files. There are two dimensions to this, namely Python version and machine architecture, with 6 Python versions and 8 architectures, making for 46 separate wheel files. (Python 3.7 has only 6 wheel files.) The 8 different architectures are required *solely* for str_util.c, since it's the only non-Python code in Urwid (or its dependencies).

This makes life a little tricky for those of us trying to build a packaged Python application to distribute to end users. Yes, the import of str_util is wrapped in a try / except, so that most of the time it will fall back to old_str_util.py if it's running on a different architecture from that which the wheel was built for, but this does not always work. Specifically, on Mac, a SyntaxError exception is raised even when using a universal2 wheel in some cases of running on an older system. That, not surprisingly, is not caught in the import try / except, and the application dies an immediate death.

My assumption is that str_util.c was created for performance reasons. That was **18 years ago**! In the intervening years, the performance of all of the systems we use today has skyrocketed compared to what we were using in 2007. So the question is, do we really still need str_util.c today? Does it really buy Urwid that much of a performance difference, in 2024, and for the majority of Urwid applications, that it's worth the complexity of having to produce wheels for all those different architectures, and the client developer complexity of having to deal with them?

Maybe it does, and it's still needed - I really don't know - but I think it's time we asked the question. If we can get rid of that one little C file, then Urwid can become a Pure Python library, allowing developers to build Pure Python applications with it. That would be awesome.

mfncooper:
Wow, this is awesome! Thanks, @penguinolog !

PR #803, opened 2024-02-15
Drop c helper and use native python.

penguinolog:
* Simplify future transition to the native strings in the canvas handling/final render
* Finally offload character width calculation to the `wcwidth`
* Normalize string width calculation and warn over incorrect offset usage
* * `calc_text_pos` and  `calc_width` over bytes are proven as unsafe
* Deprecate imports of `str_util` methods from `util`

* Normalize test_text_layout.py: drop transcoding hacks and use normal unittest logic

Fix: #801 

@jquast we need your help on `wcwidth` side to support "trim string to column" functionality since `wcwidth.wcswidth` is using private API and calling `wcwidth.wcswidth` in cycle is super expensive.

##### Checklist


urwid/str_util.py
```diff
@@ -204,17 +236,25 @@ def calc_width(text: str | bytes, start_offs: int, end_offs: int) -> int:
...
-    unis = not isinstance(text, bytes)
-    if (unis and not SAFE_ASCII_RE.match(text)) or (utfs and not SAFE_ASCII_BYTES_RE.match(text)):
-        decode = [decode_one, decode_one_uni][unis]
+    if isinstance(text, str):
+        return sum(get_char_width(char) for char in text[start_offs:end_offs])
+
+    if _byte_encoding == "utf8":
+        try:
+            return sum(get_char_width(char) for char in text[start_offs:end_offs].decode("utf-8"))
+        except UnicodeDecodeError as exc:
```
penguinolog, line 245: yes, we really sometimes hit this due to splitting in the middle of character.

urwid/str_util.py
```diff
@@ -204,17 +236,25 @@ def calc_width(text: str | bytes, start_offs: int, end_offs: int) -> int:
     if start_offs > end_offs:
         raise ValueError((start_offs, end_offs))
 
-    utfs = isinstance(text, bytes) and _byte_encoding == "utf8"
-    unis = not isinstance(text, bytes)
-    if (unis and not SAFE_ASCII_RE.match(text)) or (utfs and not SAFE_ASCII_BYTES_RE.match(text)):
-        decode = [decode_one, decode_one_uni][unis]
+    if isinstance(text, str):
+        return sum(get_char_width(char) for char in text[start_offs:end_offs])
```
penguinolog, line 240: here should be `wcwidth.wcswidth` usage, but then `calc_text_pos` may produce incompatible results.
Short details: 2 double-width calculated symbols may be rendered as single double width smile with accent (like `🤚🏻` is `🤚` + `accent`, both calculated as double width)

urwid/str_util.py
```diff
@@ -163,24 +195,24 @@ def calc_text_pos(text: str | bytes, start_offs: int, end_offs: int, pref_col: i
...
-    unis = isinstance(text, str)
-    if unis or utfs:
-        decode = [decode_one, decode_one_uni][unis]
+    if isinstance(text, str):
+        return calc_string_text_pos(text, start_offs, end_offs, pref_col)
+
+    if not isinstance(text, bytes):
+        raise TypeError(text)
+
+    if _byte_encoding == "utf8":
```
penguinolog, line 204: great place for bug if any of offsets in the middle of multibyte character

urwid/str_util.py
```diff
@@ -71,7 +76,7 @@ def decode_one(text: bytes | str, pos: int) -> tuple[int, int]:
             if lt > 3:
                 b4 = text[pos + 3]
     except Exception as e:
-        raise ValueError(f"{e}: {text=!r}, {pos=!r}, {lt=}").with_traceback(e.__traceback__) from e
```
penguinolog, line 74: syntax not supported by python 3.7

urwid/str_util.py
```diff
@@ -35,14 +36,18 @@
 _byte_encoding: Literal["utf8", "narrow", "wide"] = "narrow"
 
 
-def get_width(o: int) -> Literal[0, 1, 2]:
-    """Return the screen column width for unicode ordinal o."""
-    width = wcwidth.wcwidth(chr(o), UNICODE_VERSION)
+def get_char_width(char: str) -> Literal[0, 1, 2]:
+    width = wcwidth.wcwidth(char, UNICODE_VERSION)
     if width < 0:
```
penguinolog, line 41: control characters width calculated as `-1` (they are not expected in the string for display)

urwid/str_util.py
jquast: The solution is to use the [codecs.IncrementalDecoder()](https://docs.python.org/3/library/codecs.html#codecs.IncrementalDecoder) class.

Related, about using wcwidth, I wanted to also tell you that because of "grapheme clustering" it is better to use `wcwidth.wcswidth()` rather than `sum( ... wcwidth.wcwidth() )`, because there are character *sequences* which are measured differently than the sum of individual codepoints, for example,

```
>>> wcwidth.wcswidth('\u2665\uFE0F')
2
>>> wcwidth.wcwidth('\u2665') + wcwidth.wcwidth('\uFE0F')
1
```

However, Variation Selector 16 ([U+FE0F](https://codepoints.net/U+FE0F)) has very poor support in terminals, less than half of popular terminals support it, so I wouldn't worry about it too much! I test all popular terminals with their results here, https://ucs-detect.readthedocs.io/results.html

urwid/str_util.py
jquast: Yes, 🤚🏻 ends with [EMOJI MODIFIER FITZPATRICK TYPE-1-2](https://codepoints.net/U+1f3fb), and there is also Zero-Width Joiner Emoji Sequences ([U+200D](https://codepoints.net/U+200d)), for example,
```
seq='\U0001f9d1\U0001f3ff\U0000200d\U0001f9b3'
>>> wcwidth.wcswidth(seq)  # right
2
>>> sum(map(wcwidth.wcwidth, seq))  # wrong
4
```

jquast:
> @jquast we need your help on wcwidth side to support "trim string to column" functionality since wcwidth.wcswidth is using private API and calling wcwidth.wcswidth in cycle is super expensive.

Sorry that I'm not sure I understand this meaning, is the private API used inside wcwidth.wcswidth you are writing about, is that about its call to `_wcversion_value` ? Or maybe, the handling of wcwidth return value of -1?

---

As for the -1 workaround, and the need to use wcwidth instead of wcswidth to guard against it... the reason wcwidth functions traditionally returned -1 is that control codes like \n are indeterminate, and terminal sequences beginning with \x1b are also (like clear screen, home, move_yx, etc), and these sequences should be interpreted before sending to wcwidth functions.

But I do suggest adding new function, `wcwidth.width(control_codes='ignore')` https://github.com/jquast/wcwidth/issues/79 which would ignore single control codes, and `control_codes='replace'` to interpret (and mostly ignore) terminal sequences https://github.com/jquast/wcwidth/issues/92

Anyway I'm happy to accommodate, I am unemployed and have free time this week to help make any changes

urwid/str_util.py
penguinolog: >  The solution is to use the [codecs.IncrementalDecoder()](https://docs.python.org/3/library/codecs.html#codecs.IncrementalDecoder) class.

`start_offs` maybe already in the middle of character. I prefer to fix it permanently in the future changes with full switch of internals to the `str` as it's only one supported output for the `RawDisplay`, `CursesDisplay` in case of UTF8 mode and for web/HTML displays.

And about `wcswidth` I added comment, that we need to support compatible "trim to the column" first, or we can have render issues / hard performance issues

penguinolog:
> > @jquast we need your help on wcwidth side to support "trim string to column" functionality since wcwidth.wcswidth is using private API and calling wcwidth.wcswidth in cycle is super expensive.
> 
> Sorry that I'm not sure I understand this meaning, is the private API used inside wcwidth.wcswidth you are writing about, is that about its call to `_wcversion_value` ? Or maybe, the handling of wcwidth return value of -1?

https://github.com/jquast/wcwidth/blob/master/wcwidth/wcwidth.py#L200 is a private call.
Technically I need almost copy of `wcswidth` with extra "trim_position" argument, which will count symbols fit before this position (and return pair `width, idx`).
It will replace temporary introduced `calc_string_text_pos` (which is just optimized `str` case of `calc_text_pos`)

jquast:
> https://github.com/jquast/wcwidth/blob/master/wcwidth/wcwidth.py#L200 is a private call. Technically I need almost copy of `wcswidth` with extra "trim_position" argument, which will count symbols fit before this position (and return pair `width, idx`). It will replace temporary introduced `calc_string_text_pos` (which is just optimized `str` case of `calc_text_pos`)

Thanks I understand better now, that you wish to split a string by visible column, I will look into adding such a feature, but it will probably be a specific function, I will add an issue and reference this.

jquast (approved):
wcwidth changes look fine to me, thanks for inviting me to review

faragher:
This appears to not be functional, leading to errors such as `wcwidth() takes 1 positional argument but 2 were given`

Looking at str_util.py:40 

```width = wcwidth.wcwidth(char, UNICODE_VERSION)```

and the function in the Python3.9 package

```def wcwidth(wc, unicode_version='auto'):```

it looks like it could be repaired by editing line 40 to the following:

```width = wcwidth.wcwidth(char, unicode_version=UNICODE_VERSION)```

I apologize for not simply making the change myself, but I don't want to mess around in your repo if I don't have to.

Original reported issue with error logs: https://github.com/markqvist/Reticulum/discussions/445

Issue #496, opened 2022-01-30
Ellipsis wrap mode is supported only for UTF and limited list of other encodings


kulkarniniraj:
In case of unsupported encoding, error will happen (text layout can not be correctly built).

**Supported encodings**:
- big5
- big5hkscs
- cp874
- cp932
- cp949
- cp950
- cp1250
- cp1251
- cp1252
- cp1253
- cp1254
- cp1255
- cp1256
- cp1257
- cp1258
- euc_jp
- euc_jis_2004
- euc_jisx0213
- euc_kr
- gb2312
- gbk
- gb18030
- hz
- iso2022_jp
- iso2022_jp_1
- iso2022_jp_2
- iso2022_jp_2004
- iso2022_jp_3
- iso2022_jp_ext
- iso2022_kr
- johab
- koi8_t
- kz1048
- mac_cyrillic
- mac_greek
- mac_iceland
- mac_latin2
- mac_roman
- mac_turkish
- ptcp154
- shift_jis
- shift_jis_2004
- shift_jisx0213
- utf_32
- utf_32_be
- utf_32_le
- utf_16
- utf_16_be
- utf_16_le
- utf_7
- utf_8
- utf_8_sig

**Unsupported encodings**:
- ascii
- cp037
- cp273
- cp424
- cp437
- cp500
- cp720
- cp737
- cp775
- cp850
- cp852
- cp855
- cp856
- cp857
- cp858
- cp860
- cp861
- cp862
- cp863
- cp864
- cp865
- cp866
- cp869
- cp875
- cp1006
- cp1026
- cp1125
- cp1140
- latin_1
- iso8859_2
- iso8859_3
- iso8859_4
- iso8859_5
- iso8859_6
- iso8859_7
- iso8859_8
- iso8859_9
- iso8859_10
- iso8859_11
- iso8859_13
- iso8859_14
- iso8859_15
- iso8859_16
- koi8_r
- koi8_u

tonycpsu:
Can't reproduce this one with Urwid 2.1.2:

<img width="473" alt="image" src="https://user-images.githubusercontent.com/914142/151706071-8e3ee03a-cf8b-455d-9978-10b6bd5022f4.png">

kulkarniniraj:
Failing for me on all fonts. I am using UTF-8. Somehow it takes 3 bytes to encode '…' on my system.

penguinolog:
Can not reproduce with actual `master` branch:
```python
>>> from __future__ import annotations
[...]

PR #809, opened 2024-02-16
Fix ellipsis encoding in the text layout

penguinolog:
* Use `...` substitution in case of `…` encoding fails and cache result
* * do not encode multiple times the same value
* * handle unsupported encodings
Fix: #496

##### Checklist


AnonymouX47:
`CanvasError` is raised when `...` (triple period) is used and `1 <= maxcols < 3` e.g

```py
import urwid
urwid.set_encoding("ascii")
text = urwid.Text("Cool", wrap="ellipsis")
text.render((2,))
```
results in

```
urwid.canvas.CanvasError: Canvas text is wider than the maxcol specified:
maxcol=2
widths=[3]
text=[b'...']
urwid target encoding=ascii
```

Might also want to include these edge cases in the tests.

Thanks for your work!

PR #850, opened 2024-02-28
Cache calculation of `ListBox.rows_max()` if body supports modificati…

### LINKED ISSUE #845: [BUG] Too many items in ListBox with ScrollBar tanks performance
##### Description:

I have a scrollbar wrapped around a listbox with _lots_ of widgets ([>28745 of them](https://imgflip.com/i/8h8e5r)) which are focusable for navigation.
However, at this scale the rendering of the scrollbar/listbox starts becoming painfully slow (in the order of seconds on my Thinkpad T14), including the re-rendering on each navigational keypress.

##### Affected versions
Latest stable version from `pypi` (2.6.5)

##### Details

With some debugging, I saw that the problem seems to be in `ScrollBar.render()`, where the [size of the listbox is queried](https://github.com/urwid/urwid/blob/master/urwid/widget/scrollable.py#L494).
This causes the listbox to [iterate over all its contents](https://github.com/urwid/urwid/blob/master/urwid/widget/listbox.py#L572-L575) every time because the `size` parameter causes the caching to be avoided.

##### Possible solution (tested)

When replacing the default `urwid.ListBox` with the following implementation, navigation is _much_ smoother (after the first calculation):

```python
class CachingListBox(uw.ListBox):
    """ListBox with caching for sizes."""

    def __init__(self, body):
        """Constructor."""
        super().__init__(body)
        self._cached_sizes = {}

    def rows_max(self, size, focus):
        """Scrollable protocol for sized iterable and not wrapped around contents."""
        if size in self._cached_sizes:
            return self._cached_sizes[size]

        result = super().rows_max(size, focus)
        self._cached_sizes[size] = result
        return result
```

It wasn't possible to set only a single `self._cached_size` value, because it seems like the `rows_max` are queried with and without the scrollbar width taken into account, which basically immediately invalidates the cache.

Because the listbox content is static for us, I haven't added logic for invalidating the cache when its content changes.

If you like, I can start working on a PR for adding such caching logic to the default `ListBox` class.

max-moser:
Fixes #845 

##### Checklist
- [ ] I've branched off the `master` or `python-dual-support` branch
- [ ] I've merged fresh upstream into my branch recently


max-moser:
I noticed that this implementation doesn't work as-is because the `"modified"` signal gets sent whenever the focus changes within the listbox, causing the cache to be cleared.
This invalidates the cache on every navigational keypress, which basically negates the exact intention of the caching :sweat: 

One potential way forward would be to include information about "what has changed" in the signal/callback, but given that the callback functions don't take any arguments right now that would require a breaking change.
Adding a new signal indicating that it's explicitly the focus vs. content that's changed could also break expectations of existing implementations.

What do you think?

urwid/widget/listbox.py
```diff
@@ -365,11 +373,14 @@ def body(self, body: Iterable[Widget] | ListWalker) -> None:
             self._body = SimpleListWalker(body)
         try:
             signals.connect_signal(self._body, "modified", self._invalidate)
+            signals.connect_signal(self._body, "modified", self._clear_cache)
```
penguinolog, line 376: downside of this call is that child widgets can change their size and no notification will arrive (`_invalidate` will be called at such situation)

wardi:
Can we treat calculating the size of a ListBox as "best effort" for the purpose of drawing a scroll bar? I'm sure it doesn't really matter that the exact position is shown, especially when calculating tens of thousands of rows just to reduce that number to an integer between 0 and 24.

For example a ListWalker could estimate the relative position based on the number of widgets above/below instead of by rendering them. When there's no information available (like for ListWalker classes that lazily load data) we could render the scroll position as either top/middle/bottom based on whether there are any widgets above or below the visible area.

penguinolog:
Issue is that `ScrollBar` is rendered only if `rows` from size is less than amount of rows to  fit rendered widget.
Need to extend API to explicit render `ScrollBar` around `ListBox` or do it conditionally depends on contents length.

urwid/widget/listbox.py
max-moser: that's true, i haven't considered that aspect

max-moser:
is it safe to assume that widgets will require at least one row of space?
in this case, the check for whether the scrollbar should be rendered at all could add this as a short-circuit condition (`if len(listbox.content) > scrollbar.size[0] or listbox.rows_max > scrollbar.size[0]`).

the rendering itself could then go and estimate the listbox' actual `rows_max` by taking (random?) samples from the listbox content and extrapolating the widget sizes.
in its simplest form, this assumes a uniform distribution of the content widget sizes of course, but might be "good enough"?

max-moser:
Closing as it's fixed by #858

PR #358, opened 2019-08-29
Fix PNG screenshot race

ulidtko:
This fixes #101 and #102.

PR #867, opened 2024-04-07
Generation of PNGs screenshots: added target and misc 

isidroas:
- **Makefile: external programs detection**
- **compile_pngs.sh: export display**
- **screenshots.sh: use xdotool --sync**
- **from $c to $counter**

##### Checklist

##### Description:

### DIFF
--- a/docs/tools/compile_pngs.sh
+++ b/docs/tools/compile_pngs.sh
@@ -5,11 +5,19 @@
 DISPLAYNUM=5
 SCREENSHOTS=`dirname $0`/screenshots.sh
 
+# TODO: move this to a function which can be easily excluded and documented
 XVFB=$(which Xvfb)
 if [ -n $XVFB ]; then
 	Xvfb :$DISPLAYNUM -screen 0 1024x768x24 &
 	XVFBPID=$!
-	DISPLAY=:$DISPLAYNUM # this still doesn't work
+
+	# export DISPLAY=:$DISPLAYNUM
+	# Uncomment this if you want hide the spawned urxvt windows and only want to
+	# see the results in the pngs.
+	# It will display many warning messages in xdotool. In particular, the messages are:
+	#    XGetInputFocus returned the focused window of 1. This is likely a bug in the X server.
+	# but it works.
+
 	trap "kill $XVFBPID" EXIT
 fi
 
--- a/docs/tools/screenshots.sh
+++ b/docs/tools/screenshots.sh
@@ -14,13 +14,13 @@ RXVTWINDOWID=$(xdotool search --sync --classname "$CLASSNAME")
 export RXVTWINDOWID
 image=${1%.py}
 
-c=1
+counter=1
 while read -r line; do
 	# the echo trick is needed to expand RXVTWINDOWID variable
 	echo $line | xdotool -
 	echo "sending $line"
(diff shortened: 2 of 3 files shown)


isidroas:
Related to https://github.com/urwid/urwid/pull/358

ulidtko (commented):
Nice! Please ping me when you achieve stable results

isidroas:
This is ready for my side!

docs/tools/compile_pngs.sh
```diff
@@ -5,11 +5,19 @@
...
 	XVFBPID=$!
-	DISPLAY=:$DISPLAYNUM # this still doesn't work
+
+	# export DISPLAY=:$DISPLAYNUM
+	# Uncomment this if you want hide the spawned urxvt windows and only want to
+	# see the results in the pngs.
+	# It will display many warning messages in xdotool. In particular, the messages are:
+	#    XGetInputFocus returned the focused window of 1. This is likely a bug in the X server.
+	# but it works.
+
```
ulidtko, line 20: Sorry but no, this specific change will break building docs on CI.

We do need to explicitly point urxvt & xdotool to X11-connect to `DISPLAY=:5`. This is why we're running a "virtual X-server" on address `:5` hosted by `Xvfb` in the first place.

On a desktop linux, with your change applied, the script will inherit `DISPLAY=:0` most likely — so `urxvt` will start connecting to your graphical session, you'll see the terminal — and the `Xvfb` instance will be completely unused. Xdotool will also be able to grab screenshots. So what's the problem?

The problem is, in CI containers there's no X server and no `DISPLAY=:0`. There we need Xvfb. And to not create extra discrepancies in testing, it's best to also use Xvfb locally.

You can *comment out* the `DISPLAY=:$DISPLAYNUM` line in your local copy — but don't commit that. It will break CI.

docs/tools/screenshots.sh
```diff
@@ -10,19 +10,17 @@ urxvt -bg gray90 -b 0 +sb -fn '-misc-fixed-medium-*-*-*-*-140-*-*-*-*-*-*' \
 	-fb '-misc-fixed-bold-*-*-*-*-140-*-*-*-*-*-*' \
 	-name "$CLASSNAME" -e "$PYTHON" "$1" &
 RXVTPID=$!
-until RXVTWINDOWID=$(xdotool search --classname "$CLASSNAME"); do
-	sleep 0.1
-done
+RXVTWINDOWID=$(xdotool search --sync --classname "$CLASSNAME")
```
ulidtko, line 13: this is good :+1:

ulidtko (commented):
Good enough refactor, but does not fix the main issue with these screenshots.

@isidroas please see my comments in ~#385~, you'll also want to check pending commits in the `fix-screenshot-race` branch

Edit: I mean #358

isidroas:
> @isidroas please see my comments in \#385, you'll also want to check pending commits in the \`fix-screenshot-race\` branch

Alright, I will read them. A first question arises to me. What do you mean by `race`? Is it related to [race conditions](https://en.wikipedia.org/wiki/Race_condition#In_software) ?

> this specific change will break building docs on CI.

That change should not have effect since the `DISPLAY` variable was not being exported. I was trying to do the most minimal changes to easy reviewing.

> And to not create extra discrepancies in testing, it's best to also use Xvfb locally.

It makes sense 👍

PR #879, opened 2024-04-30
make ScrollBar check wrapped widgets for SupportsScroll (Fixes #878)

### LINKED ISSUE #878: [BUG] ListBox wrapped in AttrMap is not scrollable with ScrollBar
##### Description:

Wrapping a `ListBox` in an `AttrMap` should be transparent for operations such as adding a `ScrollBar`, but throws because the `AttrMap` is not realised to be scrollable.

- [x] `master` branch (7300acf3b1577efbacb5d54cb642aadf0849ef97)
- [x] Latest stable version from `pypi`
- [x] All versions since 2.5.0

```python

import urwid as uw
w = uw.ScrollBar(uw.AttrMap(uw.ListBox([]), {}))
```

##### Expected/actual outcome
Expected behaviour: `w` can be scrolled as if it were not wrapped in an `AttrMap`.
Actual behaviour:
```
  File "/home/koyomi/Sandbox/test_attrmap.py", line 8, in <module>
    w = uw.ScrollBar(am)
        ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/urwid/widget/scrollable.py", line 486, in __init__
    raise TypeError(f"Not a scrollable widget: {widget!r}")
TypeError: Not a scrollable widget: <AttrMap selectable box widget <ListBox selectable box widget> attr_map={}>
```

##### Cause

While `Scrollable` is aware of wrappers and descends into them to see if the original widgets are scrollable
 https://github.com/urwid/urwid/blob/7300acf3b1577efbacb5d54cb642aadf0849ef97/urwid/widget/scrollable.py#L615-L625
`ScrollBar` is not and does not
https://github.com/urwid/urwid/blob/7300acf3b1577efbacb5d54cb642aadf0849ef97/urwid/widget/scrollable.py#L484-L485

##### Fix

Make the check in `ScrollBar` analogous to the check in `Scrollable`.

rsekman:
##### Checklist

##### Description:

Fixes #878 .

### DIFF
--- a/urwid/widget/scrollable.py
+++ b/urwid/widget/scrollable.py
@@ -138,7 +138,7 @@ class SupportsRelativeScroll(WidgetProto, Protocol):
 
 
 def orig_iter(w: Widget) -> Iterator[Widget]:
-    visited = set(w)
+    visited = set((w,))
     yield w
     while hasattr(w, "original_widget"):
         w = w.original_widget
--- a/urwid/widget/scrollable.py
+++ b/urwid/widget/scrollable.py
@@ -137,6 +137,17 @@ class SupportsRelativeScroll(WidgetProto, Protocol):
     def get_visible_amount(self, size: tuple[int, int], focus: bool = False) -> int: ...
 
 
+def orig_iter(w: Widget) -> Iterator[Widget]:
+    visited = set(w)
+    yield w
+    while hasattr(w, "original_widget"):
+        w = w.original_widget
+        if w in visited:
+            break
+        visited.add(w)
+        yield w
+
+
 class Scrollable(WidgetDecoration[WrappedWidget]):
     def sizing(self) -> frozenset[Sizing]:
         return frozenset((Sizing.BOX,))
@@ -482,12 +493,6 @@ class ScrollBar(WidgetDecoration[WrappedWidget]):
         if Sizing.BOX not in widget.sizing():
             raise ValueError(f"Not a box widget: {widget!r}")
 
-        def orig_iter(w: Widget) -> Iterator[Widget]:
-            yield w
-            while hasattr(w, "original_widget"):
-                w = w.original_widget
-                yield w
-
         if not any(isinstance(w, SupportsScroll) for w in orig_iter(widget)):
             raise TypeError(f"Not a scrollable widget: {widget!r}")
 
@@ -618,12 +623,6 @@ class ScrollBar(WidgetDecoration[WrappedWidget]):
     @property
     def scrolling_base_widget(self) -> SupportsScroll | SupportsRelativeScroll:
         """Nearest `original_widget` that is compatible with the scrolling API"""
-
-        def orig_iter(w: Widget) -> Iterator[Widget]:
-            while hasattr(w, "original_widget"):
-                w = w.original_widget
-                yield w
-            yield w
 
         w = self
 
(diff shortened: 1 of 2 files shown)


urwid/widget/scrollable.py
```diff
@@ -481,7 +481,14 @@ def __init__(
         """
         if Sizing.BOX not in widget.sizing():
             raise ValueError(f"Not a box widget: {widget!r}")
-        if not isinstance(widget, SupportsScroll):
+
+        def orig_iter(w: Widget) -> Iterator[Widget]:
+            yield w
+            while hasattr(w, "original_widget"):
```
penguinolog, line 487: `w.original_widget` can point to `w`, we need to filter such behaviour to prevent infinite loop

urwid/widget/scrollable.py
```diff
@@ -481,7 +481,14 @@ def __init__(
...
+                w = w.original_widget
+                yield w
+
+        if not any(isinstance(w, SupportsScroll) for w in orig_iter(widget)):
```
penguinolog, line 491: 👍🏻

penguinolog (changes requested):

penguinolog (approved):

urwid/widget/scrollable.py
```diff
@@ -137,6 +137,17 @@ def get_first_visible_pos(self, size: tuple[int, int], focus: bool = False) -> i
...
 
+def orig_iter(w: Widget) -> Iterator[Widget]:
+    visited = set((w,))
+    yield w
+    while hasattr(w, "original_widget"):
+        w = w.original_widget
+        if w in visited:
+            break
+        visited.add(w)
+        yield w
```
wardi, line 148: http://urwid.org/reference/widget.html#urwid.WidgetDecoration.base_widget exists to get the "most" `.original_widget`

wardi: Or are there decorations that are scrollable?

rsekman: > http://urwid.org/reference/widget.html#urwid.WidgetDecoration.base_widget exists to get the "most" `.original_widget`

Looking at the code, it is implemented like this generator was originally, i.e., vulnerable to infinite loops. So the fix in https://github.com/urwid/urwid/pull/879/commits/e5377cfa2cad87836e63bdfce257931e32b09662 should be applied also in `widget_decoration.py`, at least. 

I have no idea if there are scrollable decorations; this is my first time working with the urwid code. It seems like a silly thing to implement, but if someone did they might be hit with an extremely confusing bug if we only check the `base_widget`.

urwid/widget/scrollable.py
```diff
@@ -137,6 +137,17 @@ def get_first_visible_pos(self, size: tuple[int, int], focus: bool = False) -> i
     def get_visible_amount(self, size: tuple[int, int], focus: bool = False) -> int: ...
 
 
+def orig_iter(w: Widget) -> Iterator[Widget]:
+    visited = set((w,))
```
penguinolog, line 141:
```suggestion
    visited = {w}
```

urwid/widget/scrollable.py
penguinolog: > Or are there decorations that are scrollable?

`Scrollable` is a decoration adding scroll support to the widgets:
https://github.com/urwid/urwid/blob/ba5c1c07aef2b500f6d3f1a9e2b1453f5243904c/urwid/widget/scrollable.py#L140

PR #1077, opened 2025-11-17
Refactoring: use official python way for lazy imports

penguinolog:
##### Checklist

### DIFF
--- a/urwid/__init__.py
+++ b/urwid/__init__.py
@@ -21,7 +21,7 @@
 
 from __future__ import annotations
 
-import importlib
+import importlib.util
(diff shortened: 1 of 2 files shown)


penguinolog:
> Is it intentional that this change stopped e.g. "from urwid.raw_display import Screen" or "import urwid.curses_display" from working?

It was not intensional and should be fixed as regression, since `urwid.raw_display` move was never caused `DeprecationWarning`

Proper import path is `from urwid.display.raw import Screen` or `from urwid.display import curses`, but it was added only in version 2.4.0

PR #1121, opened 2026-04-08
fix three return type annotations

evgunter:
hello! i encountered the `TextCanvas.content` issue when using urwid in my project; the other two fixes are from asking claude code to look for similar cases. let me know if there's anything else i should do to check this!

---

## Summary

- `TextCanvas.content()`: return type was missing the `list` wrapper — annotated as `Iterator[tuple[...]]` but actually yields `list[tuple[...]]` per row, consistent with all other `Canvas` subclasses (`Canvas`, `BlankCanvas`, `SolidCanvas`, `CompositeCanvas`)
- `Canvas.translate_coords()`: return type was missing `| None` on the inner data tuple, inconsistent with the declared type of `self.coords`
- `util.supports_unicode()`: `and` short-circuit can return `""` (not `bool`) when `_target_encoding` is falsy; wrapped in `bool()`

These are annotation-only fixes (except `supports_unicode` which also corrects the returned value's type). No behavioural changes.

## Test plan

- [x] `ruff format --check` passes
- [x] `ruff check` passes
- [x] `pytest` — 106 passed, 1 pre-existing failure (`_attrspec_to_escape` doctest, also fails on master)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/urwid/canvas.py
+++ b/urwid/canvas.py
@@ -458,7 +458,7 @@ class TextCanvas(Canvas):
         cols: int | None = 0,
         rows: int | None = 0,
         attr=None,
-    ) -> Iterator[tuple[object, Literal["0", "U"] | None, bytes]]:
+    ) -> Iterator[list[tuple[object, Literal["0", "U"] | None, bytes]]]:
         """
         Return the canvas content as a list of rows where each row
         is a list of (attr, cs, text) tuples.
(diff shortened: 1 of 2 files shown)


ulidtko (approved):
LOTM, thanks

penguinolog (approved):

Issue #1146, opened 2026-06-03
[Typing] Complete typing annotation of urwid


penguinolog:
## Since version 4.0.2 urwid provide provisional typing annotations.

These typing annotations are incomplete and can be not 100% correct.

#### Core

* [x] `urwid.canvas` - depends on  `urwid.text_layout`
* [x] `urwid.command_map`
* [x] `urwid.font`
* [x] `urwid.graphics`
* [x] `urwid.signals`
* [x] `urwid.split_repr`
* [x] `urwid.str_util`
* [x] `urwid.text_layout`
* [x] `urwid.util`

#### Display

Display modules depends on `urwid.canvas`

* [x] `urwid.display._posix_raw_display` - depends on `urwid.display._raw_display_base`
* [x] `urwid.display._raw_display_base` - depends on `urwid.display.common`
* [x] `urwid.display._win32`
* [x] `urwid.display._win32_raw_display`- depends on `urwid.display._raw_display_base` and `urwid.display._win32`
* [x] `urwid.display.common`
* [x] `urwid.display.curses` - depends on `urwid.display.common`
* [x] `urwid.display.escape`
* [x] `urwid.display.html_fragment` - the lowest priority
* [x] `urwid.display.lcd` - low priority
* [x] `urwid.display.raw`
* [x] `urwid.display.web`

#### Event loop

Event loop modules use `urwid.display` as control target

* [x] `urwid.event_loop.abstract_loop`
* [x] `urwid.event_loop.asyncio_loop`
* [x] `urwid.event_loop.glib_loop` - low priority (complex install, limited usage)
* [x] `urwid.event_loop.main_loop`
* [x] `urwid.event_loop.select_loop` - top priority
* [x] `urwid.event_loop.tornado_loop`
* [x] `urwid.event_loop.trio_loop`
* [x] `urwid.event_loop.twisted_loop`
* [x] `urwid.event_loop.zmq_loop` - low priority (limited support by OS)

#### Widget

All widgets are need `urwid.canvas` for proper render typing.
All text related widgets need `urwid.text_layout` for proper render typing.

##### Critical widgets

* [x] `urwid.widget.widget` - basis for all widgets and source for override warnings
* [x] `urwid.widget.text` - basis for multiple widgets
* [x] `urwid.widget.widget_decoration` - basis for widgets construction from other widgets
* [x] `urwid.widget.divider`
* [x] `urwid.widget.attr_map` - standard display attributes apply for all widgets
[...]

PR #1149, opened 2026-06-04
Typing: fully annotate `urwid.util` and partial `urwid.canvas` modules

penguinolog:
* `urwid.util`: removed unused `rle_factor` function

In mass replace `Hashable | None` with `Hashable` since `None` is `Hashable`

##### Checklist

Partial #1146

### DIFF
--- a/urwid/canvas.py
+++ b/urwid/canvas.py
@@ -41,10 +41,21 @@ from urwid.util import (
 if typing.TYPE_CHECKING:
     from collections.abc import Hashable, Iterable, Iterator, Mapping, Sequence
 
-    from typing_extensions import Literal
+    from typing_extensions import Literal, NotRequired
 
     from .display.common import AttrSpec
     from .widget import Widget
+
+    _ContentLine = list[tuple[typing.Union[AttrSpec, str, None], typing.Union[Literal["0", "U"], None], bytes]]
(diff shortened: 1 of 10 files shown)


urwid/canvas.py
```diff
@@ -273,7 +290,12 @@ def cols(self) -> int:
     def rows(self) -> int:
         raise NotImplementedError()
 
-    def content_delta(self, other: Canvas):
+    def content_delta(self, other: Canvas) -> list[int] | Iterator[_ContentLine]:
```
penguinolog, line 293: Should we deprecate it?
Looks like historical code not landed to the actual render and display logic

wardi: no objection to deprecating this

urwid/canvas.py
penguinolog, line 931: @wardi following `Display` code, this branch will cause crash since `row` is unpacked as strictly 3 items element (`for a, cs, run in row`).
Is it expected?

penguinolog: Patched locally to put `Exception` here and ran complex wizard -> branch is unreachable (following static check `content_iter` is falsy only in unit tests)

penguinolog: Will open dedicated PR

PR #1155, opened 2026-06-08
Typing: annotate `urwid.display.escape`

penguinolog:
* Prioritise UTF8 decoding against "wide" (became extremelly uncommon)
* Fix TypeError for "wide" encoding and double byte input: `str_util.within_double_byte` accepts only bytes and `chr` produce `str`. Extremely rare `if` branch due to UTF8 is default de-facto and https://peps.python.org/pep-0686 merged to python 3.15.

Partial #1146

##### Checklist

### DIFF
--- a/urwid/display/escape.py
+++ b/urwid/display/escape.py
@@ -36,6 +36,10 @@ if typing.TYPE_CHECKING:
 
     from typing_extensions import Literal
 
+    _KeyQueueData = dict[int, typing.Union[str, "_KeyQueueData"]]
+    _MouseInput = tuple[str, int, int, int]
+    _CursorPosition = tuple[typing.Literal["cursor position"], int, int]
+
 # NOTE: because of circular imports (urwid.util -> urwid.escape -> urwid.util)
 # from urwid.util import is_mouse_event -- will not work here
 import urwid.util  # isort: skip  # pylint: disable=wrong-import-position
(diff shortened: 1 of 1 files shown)


urwid/display/escape.py
```diff
@@ -504,22 +519,6 @@ def process_keyqueue(codes: Sequence[int], more_available: bool) -> tuple[list[s
...
-            code.to_bytes(1, "little"),
-            0,
-            0,
-        )
-    ):
-        if not codes[1:] and more_available:
-            raise MoreInputRequired()
-        if codes[1:] and codes[1] < 256:
-            db = chr(code) + chr(codes[1])
-            if within_double_byte(db, 0, 1):
```
penguinolog, line 520: `TypeError` on call

urwid/display/escape.py
```diff
@@ -192,46 +196,52 @@ class KeyqueueTrie:
...
+        root: _KeyQueueData,
         s: str,
         result: str,
     ) -> None:
         if not isinstance(root, MutableMapping) or not s:
             raise RuntimeError("trie conflict detected")
 
-        if ord(s[0]) in root:
-            self.add(root[ord(s[0])], s[1:], result)
+        seq_start = ord(s[0])
```
penguinolog, line 214: decode once instead of each `if`. Nano optimisation

PR #1187, opened 2026-07-14
Copy signal handlers before emission

### LINKED ISSUE #306: Signal Emission Should Copy List Of Handlers
##### Description:
If signals are deleted / added during a callback then the list of handlers mutates and you can get weird behaviour (like, some callback been called twice and others not being called at all).

- [ ] `master` branch (specify commit)
- [x] Latest stable version from `pypi` (2.0.1)
- [ ] Other (specify source)

I don't have a simple canned example, but hopefully the description is clear enough.  An easy fix would just be to make an extra list() copy.

(I don't know why an error isn't raised because you're iterating over something that is mutating, but I think it's because there's some underlying container magic).

Sanjays2402:
##### Checklist

##### Description:
Signal callbacks could mutate the live handler list during emission, skipping existing callbacks and invoking newly connected callbacks in the same dispatch. Iterate over a snapshot and add a regression test covering both disconnect and connect mutations.

Fixes #306.

### DIFF
--- a/urwid/signals.py
+++ b/urwid/signals.py
@@ -293,7 +293,7 @@ class Signals:
         """
         result = False
         handlers = getattr(obj, self._signal_attr, {}).get(name, [])
-        for _key, callback, user_arg, (weak_args, user_args) in handlers:
+        for _key, callback, user_arg, (weak_args, user_args) in handlers.copy():
             result |= self._call_callback(callback, user_arg, weak_args, user_args, args)
         return result
 
--- a/tests/test_signals.py
+++ b/tests/test_signals.py
@@ -25,6 +25,30 @@ class SiglnalsTest(unittest.TestCase):
         edit.set_edit_text("another text")
         handler.assert_not_called()
 
+    def test_handler_mutation_during_emit(self):
+        emitter = self.EmClass()
+        calls = []
+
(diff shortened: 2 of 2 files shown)


urwid/signals.py
```diff
@@ -293,7 +293,7 @@ def emit(self, obj, name: Hashable, *args: typing.Any) -> bool:
         """
         result = False
         handlers = getattr(obj, self._signal_attr, {}).get(name, [])
-        for _key, callback, user_arg, (weak_args, user_args) in handlers:
+        for _key, callback, user_arg, (weak_args, user_args) in handlers.copy():
```
penguinolog, line 296: shouldn't we use deepcopy here?

Sanjays2402: A shallow copy is intentional here. The mutation we need to isolate is append/removal on the outer `handlers` list; each handler entry and its nested argument collections are only read during dispatch. `deepcopy` would recursively copy caller-supplied `user_args`, changing object identity, and can raise for objects that do not support deep copying before their callback is invoked. `handlers.copy()` gives emission a stable iteration snapshot while preserving callback and argument identity.

penguinolog (approved):

PR #1327, opened 2026-09-18
Switch screenshots generation to 2x size and no antialiasing

penguinolog:
Downscale 50% on display

##### Checklist


wardi:
That's a big improvement, thank you.

We might still want to 2x or 3x the resolution of the images so the docs look better on high dpi displays, but that could be a separate change if you prefer.

penguinolog:
> That's a big improvement, thank you.
> 
> We might still want to 2x or 3x the resolution of the images so the docs look better on high dpi displays, but that could be a separate change if you prefer.

Looks like I'll have to do it with coming back to DejaVu font: terminus 24 font has Unicode problems

PR #1333, opened 2026-09-21
Implement synchronised output

### LINKED ISSUE #515: Feature request: support for synchronized rendering output
There is a fairly recent spec for synchronizing rendering output in the terminal. This allows the terminal to defer updates until a full "frame" can be fully rendered onscreen (subject to timeouts, so updates will not be deferred an unacceptable amount of time).  

Specs are in [this gist](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036)

The basics:
Use `CSI ? 2026 h` to enable batching output commands into a command queue.
Use `CSI ? 2026 l` when done with your current frame rendering, implicitly updating the render state by reading out the latest grid buffer state

Terminal emulator support is fairly sparse at this time (kitty, iterm2, Wezterm, and the tmux terminal multiplexer) but there appears to be interest in this feature across other terminal emulators as well (the gist lists these.)

I think raw_display.py could be modified to support this feature. 

More detail here - note this document references an older version of the standard.

https://gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec

penguinolog:
Detect support for private modes and use it by default:
* focus_reporting (1004) - override via init args (as before)
* bracketed_paste (2004) - override via init args (as before)
* synchronized_output (2026)

Fix #515

##### Checklist


urwid/display/escape.py
```diff
@@ -650,22 +650,70 @@ def process_keyqueue(
 SWITCH_TO_ALTERNATE_BUFFER = f"{ESC}[?1049h"
 RESTORE_NORMAL_BUFFER = f"{ESC}[?1049l"
 
+BRACKETED_PASTE_MODE = 2004
```
penguinolog: I'm thinking about enum since it's multiple modes and module already not short

urwid/display/_raw_display_base.py
```diff
@@ -406,22 +447,102 @@ def set_mouse_tracking(self, enable: bool = True) -> None:
...
+
+        buffer = bytearray()
+        pending = set(modes)
+        deadline = time.monotonic() + self._MODE_PROBE_TIMEOUT
+        while pending:
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                break
+            try:
+                buffer.extend(self._read_raw_input(remaining))
```
wardi: instead of blocking in this call with a "short enough" probe timeout, can we handle the terminal response as part of capturing normal input so there's no delay added at all? That would be more elegant.

urwid/display/escape.py
wardi: sure, enums are good.

